Issue #1, opened 2022-05-02
Show maximum altitude reached by drones after loading the show file


ntamas:
A common question that we get from people using the software is how they can check the maximum altitude that a drone will reach during a show. The usual response so far is to load the show in Skybrush Viewer and look at the altitude profile in the validation screen, but we should probably show the maximum altitude in the Show control panel as well.

My idea would be to put it in the secondary label of the button that allows the user to load the show file so it would become something like:

"123 drones, 4:56, max AGL 78m"

Isti115:
Idea implemented in #3.

ntamas:
Fixed in https://github.com/skybrush-io/live/commit/58e17fe4750b4ed5e3f4559e93747cf713e62ba7

PR #3, opened 2022-05-05
feat: display max AGL in loaded show file label

Isti115:
Implement #1

### DIFF
--- a/src/features/show/selectors.js
+++ b/src/features/show/selectors.js
@@ -512,10 +512,10 @@ export const getMaximumHeightInTrajectories = createSelector(
  * trajectories from the takeoff positions.
  */
 export const getProposedDistanceLimitBasedOnTrajectories = (state) => {
-  const maxHeight =
+  const maxDistance =
     getMaximumHorizontalDistanceFromTakeoffPositionInTrajectories(state);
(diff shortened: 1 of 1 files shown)


src/features/show/selectors.js
```diff
@@ -553,7 +553,9 @@ export const getShowDurationAsString = createSelector(
 export const getShowDescription = createSelector(
   getNumberOfDronesInShow,
   getShowDurationAsString,
-  (numberDrones, duration) => `${numberDrones} drones, ${duration}`
+  getMaximumHorizontalDistanceFromTakeoffPositionInTrajectories,
```
ntamas: This is the _horizontal_ distance; we need a similar selector for vertical distances and then we need to use that here

ntamas (changes requested):

src/features/show/selectors.js
ntamas: Actually, if the takeoff positions are at different altitudes, a similar selector would return an incorrect value; I guess we need a selector that returns the maximum Z coordinate across all trajectories and drones instead.

ntamas (approved):

ntamas:
Thanks a lot! :tada:

PR #12, opened 2022-07-01
Refactoring old Redux remnants into a consistent structure based on feature slices

ntamas:
Sending a PR so I can do a formal review

Summon @Isti115


ntamas:
Roughly halfway through the review (50/96 files viewed), no big issues so far, looks great!

ntamas:
All is good I think; the only significant change I've made is [this one](https://github.com/skybrush-io/live/pull/12/commits/0a42adec222bd1dbfc394dde7d5bbf19a8d82e30). I believe the original `updeep` call also used a shallow clone, but even if it didn't, I think a shallow clone is more appropriate here.

PR #13, opened 2022-07-09
fix: rotate show boundary around center of takeoff

### LINKED ISSUE #5: Rotation of show boundary with mouse gestures does not work
**Steps to reproduce**:

1. Select the yellow show boundary on the map (make sure it has a white outline).
2. Hold the Option key on macOS (I guess it's Alt on Windows), click on the boundary and drag the mouse horizontally while holding the mouse button. This should rotate the yellow boundary.
3. Release the mouse button to finish the rotation gesture. Click away from the show boundary to de-select it.
4. Note that the show boundary jumped back to the pre-rotation state.

**Expected outcome**:

The show orientation should be updated in the show settings, the show boundary should stay where it is, and the yellow takeoff markers should follow the new orientation of the show.

**Version and platform**:

Skybrush Live 2.1.0, on macOS. Probably affects other platforms as well.

Isti115:
Fixes #5

### DIFF
--- a/src/features/show/actions.js
+++ b/src/features/show/actions.js
@@ -4,6 +4,9 @@ import throttle from 'lodash-es/throttle';
 
 import { freeze } from '@reduxjs/toolkit';
 
+import { toRadians } from 'ol/math';
+import Point from 'ol/geom/Point';
+
 import { Colors } from '~/components/colors';
 import {
   addFeatureById,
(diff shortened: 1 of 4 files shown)


ntamas:
Getting there, but not quite there yet.

I've refactored the code a bit so the TransformInteractionEvent "remembers" the origin that we are rotating around, so we do not need to reach into the store at the end of the transform gesture to get the origin once again. This is a bit better, but there is still one major problem: now _all_ features on the map are rotated around the centroid of the takeoff positions (even a random polygon that you happened to draw on the map and that has nothing to do with the convex hull of the show). We must enforce the rotation origin to be the centroid of the takeoff coordinates _only_ for the convex hull of the show.

Isti115:
The situation was even worse, as trying to rotate a feature when no show was loaded straigh up crashed, but I have added the necessary condition in `TransformFeatures.js` and it works correctly now. Could you please test it again?

ntamas:
Seems good now, thanks!

PR #17, opened 2022-07-16
feat: store UAV list scroll position between modes

### LINKED ISSUE #16: Remember UAV list scroll position when toggling "Edit mapping"
Currently the scroll position jumps back to the top of the list each time mapping editing is initiated or finished.

Isti115:
Fixes #16


ntamas:
Looks good at first glance, but I have a few questions before testing this:

1. Instead of storing the scroll position in a state variable, have you considered using a ref instead? State changes trigger a re-render, so this should trigger the component to re-render every time the inner box is scrolled, and the re-render triggers an effect that updates the scroll position programmatically (to the same value nevertheless). If you used a ref, modifying the scroll position from the `onScroll` handler would not trigger a re-render.
2. This particular case of `useEffect()` seems like a good candidate for `useLayoutEffect()` instead; not sure whether it actually makes a difference, but it's worth a try. (Layout effects are executed synchronously _before_ the DOM is updated, while normal effects are executed asynchronously _after_ the DOM was already modified).
3. Can we maybe package this into a hook on its own? (Something like `useMemoizedScrollPosition()` maybe?).

Isti115:
Thank you very much for the suggestions, I have incorporated all three of them! :+1:

ntamas:
Tested with 40 virtual MAVLink drones, seems to work correctly. Thanks a lot!

PR #18, opened 2022-07-21
fix: move show file watching to be always active

### LINKED ISSUE #6: Show control panel does not warn about a modified show file on disk if it was hidden when the file was modified
This is because the `<FileWatcher>` component is mounted if and only if the "Show control panel" is visible. We need to move the component elsewhere in the component tree.

**Steps to reproduce**:

1. Open a show file
2. Authorize the start of the show in the Show control panel
3. Modify the show file on the disk
4. Deauthorize the start of the show

**Expected result**:

Skybrush should warn you that the show file was modified on the disk, but it doesn't.

Isti115:
Fixes #6


ntamas (approved):
Thanks!

Issue #9, opened 2022-06-09
Clocks LED display in a separate window


ntamas:
It would be great if we could put the clocks panel on a separate display, maybe even in full-screen mode. Investigate the best/easiest ways of doing this (separate Electron window, completely separate Electron app, or an HTML page provided by the server that can simply be opened in a browser window and enlarged to full screen).

ntamas:
This is now implemented after #24 was merged.

PR #24, opened 2022-10-24
Implement detachable panels for multi-window support (addresses #9)

Isti115:
Implement detachable panels for multi-window support (addresses #9)


src/features/detachable-panels/DetachablePanel.jsx
```diff
@@ -0,0 +1,150 @@
...
+        label={label}
+        onClick={detach}
+      />
+      {children}
+    </>
+  );
+};
+
+DetachablePanelPresentation.propTypes = {
+  attach: PropTypes.func.isRequired,
```
ntamas: Can we call these `onAttach` and `onDetach` instead? I'm slighly bothered with the apparent similarity of the `detach` and `detached` props.

src/workbench.js
```diff
@@ -52,32 +53,107 @@ const onlyWithFeature = (featureName, component) =>
...
+  },
+  'connection-list': {
+    component: views.ConnectionList,
+    label: 'Connections',
+  },
+  'dataset-list': {
+    component: views.DatasetList,
+    label: 'Datasets',
+  },
+  'dock-list': {
```
ntamas, line 74: You can safely make this detachable as well; it's a plain list-of-items just like the beacon list.

src/workbench.js
```diff
@@ -52,32 +53,107 @@ const onlyWithFeature = (featureName, component) =>
...
-  'uav-list': injectFlockFromContext(views.UAVList),
+
+export const componentRegistry = {
+  'beacon-list': {
+    component: views.BeaconList,
+    label: 'Beacons',
+    detachable: true,
+    feature: 'beacons',
+  },
+  'connection-list': {
```
ntamas, line 64: Not used any more, but you can safely make this detachable as well; it's a plain list-of-items just like the beacon list.

src/workbench.js
```diff
@@ -52,32 +53,107 @@ const onlyWithFeature = (featureName, component) =>
...
+    component: views.ConnectionList,
+    label: 'Connections',
+  },
+  'dataset-list': {
```
ntamas, line 69: Not used yet, but you can safely make this detachable as well; it's a plain list-of-items just like the beacon list.

src/workbench.js
ntamas: Actually, on second thought, let's mark it as deprecated just like we did with `messages`. I know it's not related to the feature being implemented in this PR, but let's do it anyway now that we are tidying up this part of the code.

src/workbench.js
ntamas: Actually, on second thought, let's mark it as deprecated just like we did with `messages`. I know it's not related to the feature being implemented in this PR, but let's do it anyway now that we are tidying up this part of the code.

ntamas (changes requested):
A few minor requests, otherwise I think this is good to go.

ntamas:
Thanks a lot!

Issue #15, opened 2022-07-15
Switch from WebSocket to raw TCP when Live is packaged as a desktop app


ntamas:
Skybrush Live currently uses JSON messages wrapped in a Socket.IO WebSocket connection to connect to the server. The advantage of this approach is that one can use the same connection from the desktop app and the browser-based version of Skybrush Live, and it just works. However, the server has supported raw TCP connections for quite a while now and it would be great if we could investigate what it would take to switch to TCP from WebSocket when Live is packaged in Electron.

This is likely to be a longer, non-trivial task.

The server listens for incoming TCP connections on port 5001. Live can discover this endpoint automatically with SSDP by scanning for `urn:collmot-com:service:flockwave-tcp:1` instead of `urn:collmot-com:service:flockwave-sio:1`. The connection uses newline-delimited JSON messages; in other words, you can simply split messages by looking for newline characters and each chunk must be valid JSON then. It is guaranteed that the JSON messages _themselves_ do not contain newline characters (they can only occur in strings, and in strings they are encoded as `\n`).

One thing that will surely be needed is to extend the preload script of the renderer process in Electron to provide _limited_ access to TCP functionality to the sandboxed environment in which the renderer is running. The functionality should be as limited as possible, i.e. we should not simply expose the entire Node.js `tcp` module to the renderer process but only those functions that the renderer will actually need. (Maybe the best would be that the preload module provides only a `createTCPSocket` function that returns a wrapper class around the TCP socket with nice `async` functions for reading from, writing to and closing the socket).

We will probably also need a mechanism to periodically check whether the connection is alive. WebSocket already contained this mechanism (it has special frames for that), but we need to implement it ourselves for TCP. The easiest is to set up a timer in Live to send a newline over the TCP connection every five seconds; TCP timeouts on the OS level will handle the rest if the connection is broken, the point is that there needs to be _some_ traffic over the connection even if Live has nothing to say, otherwise it is not possible to detect that the connection has been broken if the server exited without cleaning up its own end of the TCP connection.

PR #27, opened 2022-11-29
Implement TCP based communication for the Electron version (#15)

Isti115:
Implements #15

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Enable multi-window support by making panels detachable from the main layout.
 
+### Changed
+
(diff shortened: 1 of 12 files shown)


ntamas:
Looks good to me, there's only one concern that I have. Users who have already connected to a server with Socket.IO and saved the connection details in their saved state file in Skybrush Live will not be able to reconnect to the server automatically after restarting the application. More precisely, Skybrush Live will attempt to reconnect, but since the saved port is the WebSocket port, the app will try to open a raw TCP connection to the WebSocket port and the connection will eventually be closed due to protocol mismatch.

I think that we should start storing the connection protocol (TCP or WebSocket) in the saved state object, assume WebSocket when the saved state object does not contain information about the protocol, and keep on using WebSocket even from the Electron version if the saved state object still declares WebSocket (or does not declare a protocol).

PR #31, opened 2023-03-11
replace all misnamed occurrences of AGL to AHL

vasarhelyi:
Since flockwave-gps 2.3.1, flockwave-spec 1.70.0 and skybrush-server 2.10.11 the previous misnamed treatment of altitudes is finally corrected and now coordinates are expressed either as AHL (above home level, usually measured by pressure sensor etc.) or AGL (above ground level, usually measured by proximity sensors etc.). Previous references to AHL altitudes were incorrectly named as AGL due to an early Flat Earth assumption throughout the system. This pull request replaces all misnamed occurrences of AGL and ground-reference to AHL and home-reference, along with the protocol changes of the above mentioned versions of the backend.

### DIFF
--- a/src/model/sorting.js
+++ b/src/model/sorting.js
@@ -16,6 +16,7 @@ export const UAVSortKey = Object.freeze({
   GPS_FIX: 'gpsFix',
   ALTITUDE_MSL: 'amsl',
   ALTITUDE_HOME: 'ahl',
+  ALTITUDE_GROUND: 'agl',
   HEADING: 'heading',
 });
 
@@ -30,6 +31,7 @@ export const UAVSortKeys = [
   UAVSortKey.GPS_FIX,
   UAVSortKey.ALTITUDE_MSL,
   UAVSortKey.ALTITUDE_HOME,
+  UAVSortKey.ALTITUDE_GROUND,
   UAVSortKey.HEADING,
 ];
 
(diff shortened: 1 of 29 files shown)


src/model/sorting.js
```diff
@@ -15,7 +15,7 @@ export const UAVSortKey = Object.freeze({
   BATTERY: 'battery',
   GPS_FIX: 'gpsFix',
   ALTITUDE_MSL: 'amsl',
-  ALTITUDE_GROUND: 'agl',
+  ALTITUDE_HOME: 'ahl',
```
ntamas, line 18: This renaming can be a problem if the user selected AGL as the sorting key in the UAVs list in a previous version of Skybrush Live, closed the app, upgraded to this version and re-opened the app (with the old, saved sort key). It needs to be tested whether it causes a crash upon next startup or not. (Maybe a task for @Isti115 ? :)).

vasarhelyi: Maybe we can add support also for a new true AGL column and then both keys can be present

ntamas: yes, that's a possible alternative solution

vasarhelyi:
As soon as @Isti115 pushes the president font size for the altitude status widget :) I think we will be ready for review...

Isti115:
@vasarhelyi Done.

ntamas (approved):

PR #36, opened 2023-07-23
Organize, document and extend the build-time configuration options

Isti115:


config/defaults.mjs
```diff
@@ -0,0 +1,303 @@
...
+     */
+    force: false,
+  },
+
+  optimizeUIForTouch: {
+    /**
+     * Whether the UI should be adjusted primarily for touchscreen experience.
+     * @type {boolean}
+     */
+    default: undefined,
```
ntamas, line 148: What does `undefined` mean here as a default value? Shouldn't it be either `true` or `false`?

ntamas (approved):

config/defaults.mjs
isti115: Thanks for pointing this out!
Being set to `undefined` is supposed to enable the automatic detection based on the `maxTouchPoints` property of the `navigator` object, but I forgot to include that part in this PR. Also, I'll document this behavior in the jsdoc comment!

PR #39, opened 2023-12-26
Translation

ntamas:


ntamas:
@Isti115 @lengyelattil I've started working on merging this PR into the main branch, FYI.

ntamas:
For some reason I cannot update this PR; keep an eye on the `translation` branch instead, which seems to be linked to this PR but its changes are not reflected here.

Isti115:
@ntamas I would guess that the PR not updating is due to it being based on `lengyelattil:translation` instead of `skybrush-io:translation`, but despite this, visiting the branch of the organization does show this one as an active PR, so I'm not entirely sure. 🤔

PR #46, opened 2024-11-02
fix: duplicate identifier

thomas-advantitge:


ntamas (approved):

ntamas:
Thank you!

isti115:
I was wondering whether the duplication somehow occurred on my side, so I checked to see what happened, and I saw that @ntamas, you merged the `keep-flashing` branch, which was previously rebased onto my other feature branches, thus the merge included all the other changes as well. Would you be completely against a history rewrite by resetting `main` to `3946a822`? If you wouldn't allow for that, I would at least like to suggest reverting the three commits since then and cherry picking just the actual feature commit instead.

PR #47, opened 2024-11-03
New Button for check if GCS Clock is delayed based on TimeAPi

DavideLentini:
Hi , @ntamas should already know the contents of this PR. 
For now the icon of the new button is the same as the “Help” button but you can change it to your liking.
Obviously it is just a beta , and can be improved. Let me know if you may be interested in it ! thanks

### DIFF
--- a/config/baseline.ts
+++ b/config/baseline.ts
@@ -55,6 +55,7 @@ const baseline: Config = {
       'app-settings-button',
       'alert-button',
       'help-button',
+      'check-clock-sync',
       'full-screen-button',
       'session-expiry-box',
     ],
--- a/src/components/header/CheckClockSync.jsx
+++ b/src/components/header/CheckClockSync.jsx
@@ -1,0 +1,137 @@
+import config from 'config';
+import React from 'react';
+import { Translation } from 'react-i18next';
+
+import Help from '@material-ui/icons/HelpOutline';
+import GenericHeaderButton from '@skybrush/mui-components/lib/GenericHeaderButton';
+
+import store from '~/store';
+import {showNotification } from '~/features/snackbar/actions';
+import { MessageSemantics } from '~/features/snackbar/types';
+
+
+
+
+async function CheckClock(){
+
+  const MsThreShold=2; //MilliSeconds ThreShold
+  const ScThreShold=1; //Seconds ThreShold
+  const MmThreShold=1; //Minute ThreShold
+  const HhThreShold=1; //Hour ThreShold
+  const DdThreShold=1; //Days ThreShold
+
+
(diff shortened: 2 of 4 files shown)


ntamas:
@DavideLentini Instead of trying to retrieve the current timezone from the computer and asking the API to provide the current time in that timezone, shouldn't we simply retrieve the current time from the API in UTC and compare that with `Date.getTime()`, which is always in UTC? Are there any objections to do so?

DavideLentini:
@ntamas 
Yes , I think so . 
Looking at the documentation of the API that I am using in particular : 
https://timeapi.io/swagger/index.html
It seems that the json object is only returnable if a timezone or , coordinates , or an ip address is given as input .

ntamas:
True, but you can simply specify UTC as the timezone when calling the API, right?

DavideLentini:
@ntamas mmm , sorry I guess I don't really understand , 
Could you give me an example of a query with the API I shared with you?

ntamas:
```curl -X 'GET' \
  'https://timeapi.io/api/time/current/zone?timeZone=UTC' \
  -H 'accept: application/json'
```

This gives you the current time in UTC. You can then convert the `dateTime` field of the response into a UNIX timestamp and compare that with `new Date().getTIme()`.

DavideLentini:
@ntamas Ohh i get it , yes. 
So basically , I could remove all the code on the timezone , acquire the datetime from the API with this request you wrote me, 
convert it to UNIX format , and then compare it with new Date().getTIme()

DavideLentini:
@ntamas 

Then I was able to convert the API datetime to UNIX timestamps: 

```
const response = await fetch('https://timeapi.io/api/time/current/zone?timeZone=UTC',{ signal: AbortSignal.timeout(5000) });
const data = await response.json();
store.dispatch(showNotification({message:data.dateTime.toString(),semantics:MessageSemantics.SUCCESS}))
const date_time=new Date(data.dateTime).getTime(); //UNIX TIMESTAMP
store.dispatch(showNotification({message:date_time.toString(),semantics:MessageSemantics.SUCCESS}))
```
However, if I compare it to new Date().getTime()

I get a delay of 1 hour , the reason is probably that I am in UTC+1 . 
Maybe I am missing something , but shouldn't we compare our computer time with the corresponding timezone ? to see if there is any delay

DavideLentini:
@isti115 
Okay so it seems to be working! 

`const date_time=new Date(data.dateTime+“Z”).getTime();
`
If I try to change time manually , I get a delay. If you would also like to try it to give me confirmation.

I tried an API that would return a direct timestamp but to no avail.

PR #49, opened 2025-01-19
docs: Bump year in readme

darigovresearch:
Small docs change to readme


ntamas:
Thanks! I think we shouldn't need to wait for @CLAassistant so I'll just merge this as-is.

PR #51, opened 2025-01-21
Convert some files to TS

volfpeter:
Some of the changes are not used/necessary for this PR, but it was cleaner to commit them now. They will be used in future TS conversion PRs (for map-related components).

### DIFF
--- a/src/components/SwatchesColorPicker.tsx
+++ b/src/components/SwatchesColorPicker.tsx
@@ -4,7 +4,6 @@
  */
 
 import merge from 'lodash-es/merge';
-import PropTypes from 'prop-types';
 import React from 'react';
 import { SwatchesPicker } from 'react-color';
 
@@ -46,13 +45,13 @@ const hues = [
   deepOrange,
   brown,
   blueGrey,
-];
-const shades = ['700', '500', '300'];
+] as const;
+const shades = ['700', '500', '300'] as const;
 
 const pickerProps = {
-  colors: hues
-    .map((hue) => shades.map((shade) => hue[shade]))
-    .concat([['#000000', '#808080', '#FFFFFF']]),
+  colors: (
+    hues.map((hue) => shades.map((shade) => hue[shade])) as string[][]
+  ).concat([['#000000', '#808080', '#FFFFFF']]),
   width: 450,
   height: 170,
 };
@@ -64,7 +63,12 @@ const pickerStyles = {
   },
 };
 
-const SwatchesColorPicker = ({ styles, ...rest }) => (
+type SwatchesColorPickerProps = Omit<
+  React.ComponentProps<typeof SwatchesPicker>,
+  'className' | keyof typeof pickerProps
+>;
+
+export default ({ styles, ...rest }: SwatchesColorPickerProps) => (
   <SwatchesPicker
     className='borderless'
     styles={merge(pickerStyles, styles)}
@@ -72,9 +76,3 @@ const SwatchesColorPicker = ({ styles, ...rest }) => (
     {...rest}
   />
 );
-
-SwatchesColorPicker.propTypes = {
-  styles: PropTypes.object,
-};
-
-export default SwatchesColorPicker;
--- a/src/components/SwatchesColorPicker.tsx
+++ b/src/components/SwatchesColorPicker.tsx
@@ -5,7 +5,7 @@
 
 import merge from 'lodash-es/merge';
 import React from 'react';
-import { SwatchesPicker } from 'react-color';
+import { SwatchesPicker, type SwatchesPickerProps } from 'react-color';
 
 import {
   red,
@@ -49,9 +49,9 @@ const hues = [
 const shades = ['700', '500', '300'] as const;
 
 const pickerProps = {
-  colors: (
-    hues.map((hue) => shades.map((shade) => hue[shade])) as string[][]
-  ).concat([['#000000', '#808080', '#FFFFFF']]),
+  colors: hues
+    .map<string[]>((hue) => shades.map((shade) => hue[shade]))
+    .concat([['#000000', '#808080', '#FFFFFF']]),
   width: 450,
   height: 170,
 };
@@ -63,10 +63,7 @@ const pickerStyles = {
   },
 };
 
-type SwatchesColorPickerProps = Omit<
-  React.ComponentProps<typeof SwatchesPicker>,
-  'className' | keyof typeof pickerProps
->;
+type SwatchesColorPickerProps = Omit<SwatchesPickerProps, 'className'>;
 
 export default ({ styles, ...rest }: SwatchesColorPickerProps) => (
   <SwatchesPicker
--- a/src/components/SwatchesColorPicker.tsx
+++ b/src/components/SwatchesColorPicker.tsx
@@ -65,7 +65,7 @@ const pickerStyles = {
 
 type SwatchesColorPickerProps = Omit<SwatchesPickerProps, 'className'>;
 
-export default ({ styles, ...rest }: SwatchesColorPickerProps) => (
+const SwatchesColorPicker = ({ styles, ...rest }: SwatchesColorPickerProps) => (
   <SwatchesPicker
     className='borderless'
     styles={merge(pickerStyles, styles)}
@@ -73,3 +73,5 @@ export default ({ styles, ...rest }: SwatchesColorPickerProps) => (
     {...rest}
   />
 );
+
+export default SwatchesColorPicker;
--- a/src/features/beacons/selectors.ts
+++ b/src/features/beacons/selectors.ts
@@ -3,8 +3,9 @@ import { createSelector } from '@reduxjs/toolkit';
 import { globalIdToBeaconId } from '~/model/identifiers';
 import { selectionForSubset } from '~/selectors/selection';
 import type { RootState } from '~/store/reducers';
-import type { Beacon } from './types';
 import { selectOrdered } from '~/utils/collections';
+
+import type { Beacon } from './types';
 
 /**
  * Selector that calculates and caches the list of all the beacons that
--- a/src/features/beacons/selectors.ts
+++ b/src/features/beacons/selectors.ts
@@ -2,18 +2,30 @@ import { createSelector } from '@reduxjs/toolkit';
 
 import { globalIdToBeaconId } from '~/model/identifiers';
 import { selectionForSubset } from '~/selectors/selection';
-import type { RootState } from '~/store/reducers';
-import { selectOrdered } from '~/utils/collections';
+import { type AppSelector } from '~/store/reducers';
+import {
+  type Collection,
+  type Identifier,
+  selectOrdered,
+} from '~/utils/collections';
 
 import type { Beacon } from './types';
+
+/**
+ * Selector that returns the collection of beacons, that is a list of beacon IDs
+ * together with a mapping from these IDs to the information stored about them.
+ */
+export const getBeaconsAsCollection: AppSelector<Collection<Beacon>> = (
+  state
+) => state.beacons;
 
 /**
  * Selector that calculates and caches the list of all the beacons that
  * we store in the state object, in exactly the same order as they should appear
  * on the UI.
  */
-export const getBeaconsInOrder = createSelector(
-  (state: RootState) => state.beacons,
+export const getBeaconsInOrder: AppSelector<Beacon[]> = createSelector(
+  getBeaconsAsCollection,
   selectOrdered
 );
 
@@ -21,30 +33,29 @@ export const getBeaconsInOrder = createSelector(
  * Selector that calculates and caches the list of selected beacon IDs
  * from the state object.
  */
-export const getSelectedBeaconIds = selectionForSubset(globalIdToBeaconId);
+export const getSelectedBeaconIds: AppSelector<Identifier[]> =
+  selectionForSubset(globalIdToBeaconId);
 
 /**
  * Funciton that returns the name of the given beacon that should be shown to
  * the user on the UI.
  */
-export const getBeaconDisplayName = (beacon: Beacon) =>
-  (beacon ? beacon.name || beacon.id : null) || 'Unnamed beacon';
+export const getBeaconDisplayName = (beacon: Beacon): string =>
+  beacon?.name ?? beacon?.id ?? 'Unnamed beacon';
 
 /**
  * Selector that returns the IDs of all the beacons that have no basic
  * information fetched from the server yet.
  */
-export const getBeaconIdsWithoutBasicInformation = createSelector(
-  (state: RootState) => state.beacons,
-  (beacons) => {
+export const getBeaconIdsWithoutBasicInformation: AppSelector<Identifier[]> =
+  createSelector(getBeaconsAsCollection, (beacons) => {
     const result = [];
 
     for (const [key, value] of Object.entries(beacons.byId)) {
-      if (typeof value.name === 'undefined') {
+      if (value.name === undefined) {
         result.push(key);
       }
     }
 
     return result;
-  }
-);
+  });
--- a/src/features/beacons/selectors.ts
+++ b/src/features/beacons/selectors.ts
@@ -2,6 +2,8 @@ import { createSelector } from '@reduxjs/toolkit';
 
 import { globalIdToBeaconId } from '~/model/identifiers';
 import { selectionForSubset } from '~/selectors/selection';
+import type { RootState } from '~/store/reducers';
+import type { Beacon } from './types';
 import { selectOrdered } from '~/utils/collections';
 
 /**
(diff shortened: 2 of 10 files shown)


src/components/SwatchesColorPicker.tsx
```diff
@@ -46,13 +45,13 @@ const hues = [
...
+] as const;
+const shades = ['700', '500', '300'] as const;
 
 const pickerProps = {
-  colors: hues
-    .map((hue) => shades.map((shade) => hue[shade]))
-    .concat([['#000000', '#808080', '#FFFFFF']]),
+  colors: (
+    hues.map((hue) => shades.map((shade) => hue[shade])) as string[][]
+  ).concat([['#000000', '#808080', '#FFFFFF']]),
```
isti115: IMO we should provide an explicit type parameter to `map` instead of forcing the assertion:
```js
  colors: hues
    .map<string[]>((hue) => shades.map((shade) => hue[shade])))
    .concat([['#000000', '#808080', '#FFFFFF']]),
```

src/components/SwatchesColorPicker.tsx
```diff
@@ -64,17 +63,16 @@ const pickerStyles = {
   },
 };
 
-const SwatchesColorPicker = ({ styles, ...rest }) => (
+type SwatchesColorPickerProps = Omit<
+  React.ComponentProps<typeof SwatchesPicker>,
+  'className' | keyof typeof pickerProps
+>;
```
isti115: Why don't we use the `SwatchesPickerProps` type provided by `@types/react-color`? (Is re-extracting the type using `React.ComponentProps` more conventional? I don't have much experience with the UI related best practices yet.) Also, I personally wouldn't exclude the possibility to override parts of `pickerProps` through the `...rest` props.

src/components/SwatchesColorPicker.tsx
```diff
@@ -64,17 +63,16 @@ const pickerStyles = {
...
+
+export default ({ styles, ...rest }: SwatchesColorPickerProps) => (
   <SwatchesPicker
     className='borderless'
     styles={merge(pickerStyles, styles)}
     {...pickerProps}
     {...rest}
   />
 );
```
isti115: Isn't the linter usually angry about default exporting an unnamed arrow function?

---

_Edit:_ It is. https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-anonymous-default-export.md

I'm a bit conflicted about this rule, as sometimes I feel like stating the name of the component again is redundant, but so far we have conformed to the suggestion, and I guess that it is worth sticking to it for the sake of consistency. (Or if we consider it unimportant, let's explicitly disable the rule!)

src/features/settings/selectors.d.ts
```diff
@@ -1 +1,4 @@
+import type { APIKeysRecord } from '~/APIKeys';
+
 export function shouldOptimizeUIForTouch(state: any): boolean;
+export function getAPIKeys(state: any): APIKeysRecord;
```
isti115: Let's minimize the usage of `any`, we have a proper type (`RootState`) for the state in `~/src/store/reducers.ts`.

src/features/mission/actions.js
```diff
@@ -8,7 +8,6 @@ import unary from 'lodash-es/unary';
 import { getDistance as haversineDistance } from 'ol/sphere';
 
 import { findAssignmentInDistanceMatrix } from '~/algorithms/matching';
-import { Colors } from '~/components/colors';
```
isti115, line 11: I guess that this was removed because of being unnecessarily imported. In this case let's also get rid of `LabelStyle`, which seems to be superfluous as well.

src/features/beacons/selectors.ts
```diff
@@ -2,6 +2,8 @@ import { createSelector } from '@reduxjs/toolkit';
 
 import { globalIdToBeaconId } from '~/model/identifiers';
 import { selectionForSubset } from '~/selectors/selection';
+import type { RootState } from '~/store/reducers';
+import type { Beacon } from './types';
```
isti115, line 12: Let's put the relative import into a new paragraph below separated by an empty line, so it doesn't blend in with the rest!

src/features/beacons/selectors.ts
```diff
@@ -24,15 +26,15 @@ export const getSelectedBeaconIds = selectionForSubset(globalIdToBeaconId);
  * Funciton that returns the name of the given beacon that should be shown to
  * the user on the UI.
  */
-export const getBeaconDisplayName = (beacon) =>
+export const getBeaconDisplayName = (beacon: Beacon) =>
```
isti115: This function is not actually a selector, let's move it to `./utils.ts`!
Also, I appreciate explicit return type annotations.

src/features/docks/actions.ts
```diff
@@ -5,10 +5,10 @@ import { dockIdToGlobalId } from '~/model/identifiers';
...
  *
- * @param {Array.<string>} ids  the IDs of the selected docking stations.
+ * @param ids  The IDs of the selected docking stations.
  *        Any docking stations whose ID is not in this set will be deselected,
  *        and so will be any feature that is not a docking station.
- * @return {Object} an appropriately constructed action
+ * @return An appropriately constructed action.
  */
-export const setSelectedDockIds = (ids) =>
+export const setSelectedDockIds = (ids: string[]) =>
```
isti115: I would prefer to use the type `Identifier` from `~/utils/collections` here, which is just an alias for `string`, but adds some semantic meaning to the type.

src/features/beacons/selectors.ts
```diff
@@ -10,7 +12,7 @@ import { selectOrdered } from '~/utils/collections';
  * on the UI.
  */
 export const getBeaconsInOrder = createSelector(
-  (state) => state.beacons,
+  (state: RootState) => state.beacons,
```
isti115: Instead of annotating these sub-selectors inline, I would prefer creating a properly typed reusable selectors, such as:
```ts
export const getBeaconsAsCollection: AppSelector<Collection<Beacon>> = (
  state
) => state.beacons;
```

src/features/beacons/selectors.ts
```diff
@@ -10,7 +12,7 @@ import { selectOrdered } from '~/utils/collections';
  * on the UI.
  */
 export const getBeaconsInOrder = createSelector(
```
isti115: Even though this might seem overzealous, I'd still annotate this as `AppSelector<Beacon[]>`, which I would much rather encounter later on instead of the one inferred by TypeScript:
```ts
const getBeaconsInOrder: ((state: EmptyObject & {
    alert: AlertSliceState;
    beacons: BeaconsSliceState;
    clocks: ClocksSliceState;
    connections: ConnectionsSliceState;
    ... 36 more ...;
    workbench: WorkbenchSliceState;
}) => Beacon[]) & OutputSelectorFields<...> & {
    ...;
}
```

src/features/beacons/selectors.ts
```diff
@@ -24,15 +26,15 @@ export const getSelectedBeaconIds = selectionForSubset(globalIdToBeaconId);
...
  */
-export const getBeaconDisplayName = (beacon) =>
+export const getBeaconDisplayName = (beacon: Beacon) =>
   (beacon ? beacon.name || beacon.id : null) || 'Unnamed beacon';
 
 /**
  * Selector that returns the IDs of all the beacons that have no basic
  * information fetched from the server yet.
  */
 export const getBeaconIdsWithoutBasicInformation = createSelector(
```
isti115: Similarly to above, this time with `AppSelector<Identifier[]>`.

src/features/beacons/selectors.ts
```diff
@@ -24,15 +26,15 @@ export const getSelectedBeaconIds = selectionForSubset(globalIdToBeaconId);
...
-  (state) => state.beacons,
+  (state: RootState) => state.beacons,
```
isti115: Similarly to above, reuse `getBeaconsAsCollection`.

src/features/docks/selectors.ts
```diff
@@ -10,7 +11,7 @@ import { selectOrdered } from '~/utils/collections';
  * on the UI.
  */
 export const getDocksInOrder = createSelector(
-  (state) => state.docks,
+  (state: RootState) => state.docks,
```
isti115, line 14: Similarly to beacons, I would introduce a `getDocksAsCollection` selector and explicitly annotate all other selectors as well.

src/features/beacons/selectors.ts
isti115, line 1: Since I've left so many comment on this file, I thought that I'd attach a version in which all of them are addressed and the suggestions are applied, furthermore two linter errors are fixed:
- @typescript-eslint/prefer-nullish-coalescing : Prefer using nullish coalescing operator (`??`) instead of a logical or (`||`), as it is a safer operator.
- unicorn/no-typeof-undefined : Compare with `undefined` directly instead of using `typeof`.

```ts
import { createSelector } from '@reduxjs/toolkit';

import { globalIdToBeaconId } from '~/model/identifiers';
import { selectionForSubset } from '~/selectors/selection';
import { type AppSelector } from '~/store/reducers';
import {
  type Collection,
  type Identifier,
  selectOrdered,
} from '~/utils/collections';

import type { Beacon } from './types';

/**
 * Selector that returns the collection of beacons, that is a list of beacon IDs
 * together with a mapping from these IDs to the information stored about them.
 */
export const getBeaconsAsCollection: AppSelector<Collection<Beacon>> = (
  state
) => state.beacons;

/**
 * Selector that calculates and caches the list of all the beacons that
 * we store in the state object, in exactly the same order as they should appear
 * on the UI.
 */
export const getBeaconsInOrder: AppSelector<Beacon[]> = createSelector(
  getBeaconsAsCollection,
  selectOrdered
);

/**
 * Selector that calculates and caches the list of selected beacon IDs
 * from the state object.
 */
export const getSelectedBeaconIds: AppSelector<Identifier[]> =
  selectionForSubset(globalIdToBeaconId);

/**
 * Funciton that returns the name of the given beacon that should be shown to
 * the user on the UI.
 */
export const getBeaconDisplayName = (beacon: Beacon): string =>
  beacon?.name ?? beacon?.id ?? 'Unnamed beacon';

/**
 * Selector that returns the IDs of all the beacons that have no basic
 * information fetched from the server yet.
 */
export const getBeaconIdsWithoutBasicInformation: AppSelector<Identifier[]> =
  createSelector(getBeaconsAsCollection, (beacons) => {
    const result = [];

    for (const [key, value] of Object.entries(beacons.byId)) {
      if (value.name === undefined) {
        result.push(key);
      }
    }

    return result;
  });
```

src/features/beacons/selectors.ts
```diff
@@ -24,15 +26,15 @@ export const getSelectedBeaconIds = selectionForSubset(globalIdToBeaconId);
...
   (beacons) => {
     const result = [];
```
isti115: This really looks to me like it should just be a `filter` followed by a `map`:
```ts
export const getBeaconIdsWithoutBasicInformation: AppSelector<Identifier[]> =
  createSelector(getBeaconsAsCollection, (beacons) =>
    Object.entries(beacons.byId)
      .filter(([_key, value]) => value.name === undefined)
      .map(([key, _value]) => key)
  );
```

isti115 (changes requested):
Sorry for the copious amount of comments. :upside_down_face:
There might indeed be some truth to the following recurring meme:

image

src/components/SwatchesColorPicker.tsx
volfpeter: Done in 5c32a062c5912da324907d6065569d7a754efee8

src/components/SwatchesColorPicker.tsx
volfpeter: I imported the type as well. It's cleaner, the old version sometimes feel more convenient (and actually I didn't check whether the lib's typing actually exports the prop type). Omitting `keyof pickerProps` was just to be strict, but it's removed now. See 5c32a062c5912da324907d6065569d7a754efee8

src/components/SwatchesColorPicker.tsx
volfpeter: If I'm honest, I don't like using default exports, in my own project I only use them in very few cases (e.g. exporting the reducer from a slice definition). In these cases I like not having a redundant named variable.

Where is the eslint config? :slightly_smiling_face: A CI job would also be handy so similar issues can be found without a review.

In any case, I'll change it. (Done)

src/features/settings/selectors.d.ts
volfpeter: Thanks, I forgot about this... Fixed.

src/features/mission/actions.js
volfpeter: Good catch, I removed that as well.

src/features/beacons/selectors.ts
volfpeter: Done. Couldn't prettier sort the imports? I got so used to `ruff` doing all of this... (note: in general, import blocks are a bit mixed up in quite few files). (Done)

src/features/beacons/selectors.ts
volfpeter: I can do that, but it's not really what this PR is about. I think I saw several selectors like this.
I'm just curious: what's the benefit of explicit return type annotations? I always add them in Python, but typing and tooling is quite different there; I never used it in TS in previous projects.

src/features/docks/actions.ts
volfpeter: Done.

src/features/beacons/selectors.ts
volfpeter: I committed it, although personally I feel this `AppSelector`-based typing is very verbose.

src/features/beacons/selectors.ts
volfpeter: Feel free to commit the improvement, but this comment is also not related to the changes in the PR.

volfpeter:
@isti115 Thanks for the comments (I thought I was nitpicky :slightly_smiling_face:). I responded to almost everything and fixed what I felt is reasonable. Unfortunately it seems we have a very different approach to typing (e.g. things like `AppSelector` and type inference). I hope the current state of the PR will be okay for you.

src/components/SwatchesColorPicker.tsx
isti115: The eslint config is in the `package.json` file. (Although it's a bit disguised under the `xo` key. I have already considered suggesting a switch to vanilla eslint with `eslint-config-xo` before to take advantage of `eslint_d`, but so far decided not to bikeshed too much over this.)

src/features/beacons/selectors.ts
isti115: I'm not sure whether prettier has such functionality, I usually rely on the _"Organize imports"_ code action provided by the TypeScript language server for this task.

src/features/beacons/selectors.ts
isti115: Well, for one I think that it's nice, as it gives a reaffirmation that someone had consciously and deliberately decided what the return type should be and it didn't just accidentally end up being what it is. Another reason might be that LSP based type inference is not available everywhere, for example while reviewing this PR on GitHub I cannot just enable type hints to see what a given function would return. If it's not explicitly annotated I sometimes have to either think it through in my head, which is error prone, or alternatively open it in my development environment, which requires a context change and isn't available on my phone.

src/components/SwatchesColorPicker.tsx
volfpeter: Interesting, it did not flag this for me.

isti115 (approved):
I did indeed get a bit carried away, but once I had the momentum, I couldn't resist it enough to stop halfway. 😀 My more serious concerns (forced `as` type assertion and `any`) have absolutely been addressed, thanks for taking all the feedback! 🙂

Regarding the comments, that were seemingly unrelated to the PR: I think they are about as related as removing unused imports from a random `.js` file in a PR titled "Convert some files to TS". 😉

@ntamas told me once that he likes to take a _"good little scout"_ approach and tries to leave everything behind in a bit nicer shape than it was before he touched a given file. Since we're already adding four more lines to `src/features/settings/selectors.d.ts` - which only had one so far - I would probably fix the typing of that remaining single line as well, but that might also be heavily biased by my fear towards introducing occurrences of `any` into the codebase. I take every chance I can for eliminating them.

(I can also see how doing every little unrelated thing at the same time can also lead to chaos. I have several such changes stashed in my local git repo for maybe creating a generic code quality improvement commit later, but haven't gotten around to doing it yet...)

volfpeter:
> "good little scout"

Yeah, everyone learns this very quickly from Tamás. There's no other way of doing SW, although you must always draw the line somewhere, otherwise you'll keep polishing the same fireplace forever.

Anyway, there was plenty of new info I can take into my next PR (hint: a bunch of map-related refactoring and TS conversion, not sure I'll ask for a review from you though :slightly_smiling_face:  )

ntamas:
LGTM. :-D

(After the review from @isti115 I feel there's nothing else to add).

ntamas (approved):

PR #52, opened 2025-01-21
Refactor map related components

volfpeter:
- Extract many map-related components to `map/components`.
- Convert many components to TS.
- Add support for layer exclusion.


src/components/map/style.ts
ntamas, line 1: This file is probably unused or not used much. Its contents is specific to styling map tiles from [Mapbox](https://www.mapbox.com/) so I'd rather keep the name as is (i.e.` mapbox.ts` or `mapbox-style.ts`).

ntamas (approved):
I'll merge this and then do the rename that I mentioned in the review as a separate commit

PR #53, opened 2025-01-29
Move more map related components to the components package

volfpeter:
And some TS conversion


volfpeter:
Argh, wrong target branch. Closing.

PR #54, opened 2025-01-29
Move more map related components

volfpeter:
... to the components package. And some TS conversion.


src/components/map/interactions/TransformFeatures.js
```diff
@@ -12,15 +12,14 @@ import PropTypes from 'prop-types';
...
 import { getCenterOfFirstPointsOfTrajectoriesInWorldCoordinates } from '~/features/show/selectors';
 import {
-  isOriginId,
-  globalIdToAreaId,
   CONVEX_HULL_AREA_ID,
+  globalIdToAreaId,
+  isOriginId,
 } from '~/model/identifiers';
-import store from '~/store';
+import store from '~/store'; // TODO(vp): it would be nice to not have to import the store here.
```
ntamas, line 22: Just to add some context as to why the `store` is imported in case you can come up with a better solution: we need the store in `TransformFeaturesInteraction` because of only one thing. When the user attempts to rotate one or more features on the map, we rotate it around the center of the extent of the feature(s) being rotated because this is the most intuitive to users. However, there is one exception: if the last selected feature is the polygon that represents the convex hull of the show, we rotate this around the mean of the coordinates of the initial points of the trajectories, and in order to get this point we need to reach out to the store.

One possible solution that I can think of is to provide an optional callback prop to the `TransformFeaturesInteraction` that receives the current selection and returns the center of the rotation interaction. This callback can then be injected from the point where the `<TransformFeaturesInteraction>` component is initialized where we probably have better access to the store itself (possibly in a connected component). When no such callback is provided, we fall back to the behaviour that selects the center of the extent of all selected features, which works in 99% of the cases by default.

src/components/map/mapbox-style.ts
```diff
@@ -31,7 +31,7 @@ function createMapboxStreetsV6Style() {
             'https://unpkg.com/@mapbox/maki@4.0.0/icons/' +
             iconName +
             '-15.svg',
-          // TODO(volfpeter): Is this correct (copied from views.map.styles.mapbox)?
+          // TODO(vp): Is this correct (copied from views.map.styles.mapbox)?
           // Seems like the imgSize prop does not exist.
```
ntamas, line 35: `imgSize` was removed in OpenLayers 8 so we can get rid of this as well. See more details [here](https://github.com/openlayers/openlayers/blob/main/changelog/upgrade-notes.md#removal-of-olstyleicons-imgsize-property).

ntamas (approved):

src/components/map/interactions/TransformFeatures.js
volfpeter: Thanks for the detailed explanation! Do you mind if I copy your comment to the code to persist it? At least the first half of it.

Having worked with Live for only a couple of weeks, I'm definitely no expert here, but I agree, dependency injection is a pretty straightforward a clean solution. (I'd maybe make the callback return the center of rotation or `undefined`, and then handle `undefined` by using default method, but this is low level detail).

ntamas: Sure, go ahead and copy it to the code.

PR #57, opened 2025-04-10
Site survey

volfpeter:
Feature/dialog name to be decided.

TODOs:

- [x] Don't call show-adapt if the user switches between the config and review stages without changing the adapt parameters. Basic version is OK for now (if a form item changes once, drop the stored adapted show).
- [ ] Don't let the user open the dialog if there's no connection to the server.
- [ ] Restore map controls and disconnect them from map view's map.


src/components/map/layers/ShowInfoLayer.tsx
```diff
@@ -28,19 +29,27 @@ import {
   whiteVeryThinOutline,
 } from '~/utils/styles';
 
-const globalIds = {
-  convexHull: areaIdToGlobalId(CONVEX_HULL_AREA_ID),
-};
-
 // === Convex hull ===
 
+export enum ConvexHullVariant {
```
volfpeter, line 34: I know Gábor uses these terms, but I must admit I don't like them, which I why I avoided using them (especially as global IDs). They are not descriptive at all, and they also don't generalize for other potential use-cases, which is not good for a brand new feature. FWIW: I don't need we need separate global IDs now at all.

isti115: Hmm, interestingly enough I think that they pretty nicely convey their meanings. (I might be used to them from the generic mission context, where they were already introduced a few years ago to be able to resume an interrupted mission, in which case only the useful segments needed to be considered without the ancillary parts.)

I don't necessarily insist on using separate global IDs, but to me they represent two conceptually different features, which could potentially even appear on the same map at some later stage, in which case they would actually be required to have their own respective IDs anyway. If that would sit better with you, we could keep the previous single `CONVEX_HULL_AREA_ID`  constant and form the IDs by appending the variant at the end. Still, we need to be able to differentiate between the two to have them behave differently during the transformation interaction.

volfpeter: > I think that they pretty nicely convey their meanings. (I might be used to them from the generic mission context

Interesting, being so new here, I obviously wasn't aware of that. It makes more sense this way, although I think in both cases the net/gross show definition needs detailed explanation (see your explanation :)). But as I said, since it's already in use for other purposes, it's okay for me to keep it.

> we could keep the previous single `CONVEX_HULL_AREA_ID` constant and form the IDs by appending the variant at the end. Still, we need to be able to differentiate between the two to have them behave differently during the transformation interaction.

I was about to recommend exactly this :slightly_smiling_face: I don't think global ID constants scale too well. Based on Gábor's comments, we may need 5+ soon.

src/views/show-control/SiteSurveyButton.tsx
```diff
@@ -0,0 +1,99 @@
+import ListItem from '@material-ui/core/ListItem';
+import ListItemText from '@material-ui/core/ListItemText';
+import React, { useCallback } from 'react';
+import { useTranslation } from 'react-i18next';
+import { connect } from 'react-redux';
+
+// @ts-ignore
```
ntamas: I don't think we need `@ts-ignore` here. Typings are provded in `types/` and the TS compiler and the linter should pick them up (it is doing so for me in VS Code).

src/flockwave/queries.js
```diff
@@ -14,6 +14,40 @@ import { errorToString } from '~/error-handling';
...
+    {
+      type: 'X-SHOW-ADAPT',
+      show,
+      transformations,
+      environment: {
+        location: {
+          origin: [
+            // Convert the origin to a lat-lon integer pair.
+            // Unit must be 1e-7 degrees (so multiple by 1e7).
+            Math.round(coordinateSystem.origin[1] * 1e7),
```
ntamas: We have a `toScaledJSONFromLonLat()` function for this type of conversion so I will just use that here.

ntamas (approved):

volfpeter:
Thanks Tamas!

PR #58, opened 2025-04-13
Fly here confirmation dialog

thomas-advantitge:


ntamas:
Conditional on merging https://github.com/skybrush-io/skybrush-server/pull/9/

PR #65, opened 2025-05-03
Channel indicator & hotkey

thomas-advantitge:
Adds a little indicator (number "1" or "2") to show the active channel without having to open the menu.
Also adds a shortcut to toggle between the channels.

image

### DIFF
--- a/src/components/header/ChannelIndicator.jsx
+++ b/src/components/header/ChannelIndicator.jsx
@@ -1,0 +1,36 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { connect } from 'react-redux';
+import { makeStyles } from '@material-ui/core/styles';
+
+import { getPreferredCommunicationChannelIndex } from '~/features/mission/selectors';
+
+const useStyles = makeStyles(() => ({
+  root: {
+    position: 'absolute',
+    bottom: 5,
+    right: 5,
+    color: 'white',
+    width: 16,
+    height: 16,
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    fontSize: 11,
+  },
+}));
+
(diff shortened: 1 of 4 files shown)


ntamas:
Wow, thanks a lot, this seems useful!

@vasarhelyi Are you okay with this? Especially with adding the channel switch to the `c` hotkey, i.e. a hotkey without modifiers. I'm slightly leaning towards `mod+c` instead, but that would be `Ctrl-C` on Windows and `Cmd-C` on Mac, overwriting the shortcut for the "copy" operation so I don't really like that either.

vasarhelyi:
Some thoughts if you asked me:
- I would move the channel number a few pixels to the right
- Ctrl-C is very "universal" and commonly used, please do not overwrite that!
- The hotkeys without modifiers now are only `w` and `s`, we can assign `c` to channel without interference. This is not a very harmful command, and pressing it again is basically "undo" so easy to cancel it.
- Idea: we could also consider modifying Ctrl-B to simple `b` then to have similar complexity for "broadcast" mode (I know your answer Tamás, backward compatibility :) ).

ntamas:
`B` can be added alongside `Ctrl-B` without breaking backward compatibility if we want to go down this road (i.e. more command hotkeys without a specific modifier). I would definitely keep modifiers for "destructive" actions (actions that cannot easily be undone, like a takeoff or land command). Not having a modifier here is probably okay -- as you said, it is easy to revert.

ntamas:
Thanks a lot!

ntamas:
I'll make one small tweak: I will _hide_ the "1" indicator if the primary channel is in use but keep "2" to indicate the secondary channel. This is because in the typical and recommended Skybrush Live + Sidekick setup the server only handles one channel, and in this case it's misleading to call attention to the fact that it is the primary channel.

PR #66, opened 2025-05-03
UAV status panel

thomas-advantitge:
Adds a panel for the UAV statuses which is useful for show recordings as now this is only visible when hovering over the UAV status summary header item.
image

### DIFF
--- a/src/components/uavs/UAVStatusSummary.jsx
+++ b/src/components/uavs/UAVStatusSummary.jsx
@@ -160,7 +160,7 @@ const showUAVsList = (workbench, dispatch) => () => {
     return;
   }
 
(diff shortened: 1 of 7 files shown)


ntamas:
Thanks for the contribution! This was implemented, albeit in a slightly different manner, see commit 4b5f8d55 and a few commits before that. You can now hover with the mouse over one of the header widgets to make it show the tooltip, then click the "Pin" icon in the tooltip to pin that tooltip to the workbench as a separate panel.

This is now implemented for the UAV status, the RTK status and the selection groups header widget, but since the mechanism is generic, it is easy to add this to any of the remaining header tooltips if needed.

PR #67, opened 2025-05-07
Fix selection in lists using the keyboard

volfpeter:
It seems there were multiple issues with the existing code:

- Range selection (Shift + direction) did not work, there was a TODO for it.
- Selection with PgUp/PgDn/Home/End did not work if the naively calculated new index was outside of `[0, numOfItems - 1]` interval.
- Selection with Shift+direction didn't work when there was no selected item.

I tested the fix on the UAVs list (both with grid and list layout), I think everything works correctly, but maybe there are supported combinations I didn't think of. @vasarhelyi Could you please test this?


ntamas (approved):

ntamas:
LGTM. Thanks!

PR #68, opened 2025-05-09
feat: show drones in the site survey dialog

volfpeter:


src/features/site-survey/selectors.ts
```diff
@@ -217,3 +231,51 @@ export const selectApproximateConvexHullOfFullShowInWorldCoordinates =
...
+      LayerType.TILE_SERVER,
+      LayerType.FEATURES,
+    ]);
+    // The index of the last base layer, that should be below the UAVs layer.
+    const targetIndex = result.findLastIndex((l) => targetLayers.has(l.type));
+    // Insert the UAVs layer at the right place.
+    result.splice(targetIndex + 1, 0, uavsLayer);
+    return result;
+  }
+);
```
isti115, line 281: Personally I would find the following mutation-free version much easier to understand and maintain:
```typescript
export const getVisibleLayersInOrder = createSelector(
  _getVisibleLayersInOrder,
  selectDronesVisible,
  (layers, dronesVisible): Layer[] => {
    // If drones are not visible, we can return the filtered layers array.
    if (!dronesVisible) {
      return layers.filter((l) => l.type !== LayerType.UAVS);
    }

    // UAVs layer to be used with a default if `layers` doesn't contain one.
    const uavsLayer: Layer = layers.find((l) => l.type === LayerType.UAVS) ?? {
      id: 'site-survey-uavs',
      type: LayerType.UAVS,
      label: 'UAVs',
      visible: true,
      parameters: {},
    };

    // Some of the base layers that should be below the UAVs layer.
    const targetLayers = new Set([
      LayerType.BASE,
      LayerType.GRATICULE,
      LayerType.TILE_SERVER,
      LayerType.FEATURES,
    ]);

    // The index of the last base layer, that should be below the UAVs layer.
    const targetIndex = layers.findLastIndex((l) => targetLayers.has(l.type));
    return layers.toSpliced(targetIndex + 1, 0, uavsLayer);
  }
);
```

volfpeter: I don't think that code is correct, we could end up with two UAVs layers.

_(My original version was very similar to this, and even though the arrays are short, I expected a comment about too many iterations over it :slightly_smiling_face: I just noticed that I still left an extra one in that I should have removed. Anyway, I agree simple is better than complex, so if you have no complaints about this, I'll happy simplify the code. )_

ntamas: Personally, I'm okay with the current implementation even if it is not free of mutations (we are mutating an array that was created in the same function so it's OK to me).

isti115: > _"we could end up with two UAVs layers"_

Ouch, sorry, you're right! :man_facepalming: Looking at the edit history of my comment there has even been an extra line at the bottom originally that said:
> _"(It is a bit less efficient by potentially iterating the layers array one more time compared to the original, but that really shouldn't matter in this case...)"_

Almost as if I have actually thought about needing the filter anyway, but somehow didn't include it in the actual code... :upside_down_face:

---
_Edit:_ A (now hopefully) correct version of my suggestion, for posterity:
```typescript
export const getVisibleLayersInOrder = createSelector(
  _getVisibleLayersInOrder,
  selectDronesVisible,
  (layers, dronesVisible): Layer[] => {
    // UAVs layer to be used with a default if `layers` doesn't contain one.
    const uavsLayer: Layer = layers.find((l) => l.type === LayerType.UAVS) ?? {
      id: 'site-survey-uavs',
      type: LayerType.UAVS,
      label: 'UAVs',
      visible: true,
      parameters: {},
    };

    // All the visible layers except UAVs.
    const otherLayers = layers.filter((l) => l.type !== LayerType.UAVS);

    // If drones should not be visible, we can return the filtered layers array.
    if (!dronesVisible) {
      return otherLayers;
    }

    // Some of the base layers that should be below the UAVs layer.
    const targetLayers = new Set([
      LayerType.BASE,
      LayerType.GRATICULE,
      LayerType.TILE_SERVER,
      LayerType.FEATURES,
    ]);

    // The index of the last base layer, that should be below the UAVs layer.
    const targetIndex = otherLayers.findLastIndex((l) =>
      targetLayers.has(l.type)
    );
    return otherLayers.toSpliced(targetIndex + 1, 0, uavsLayer);
  }
);
```

PR #69, opened 2025-05-13
Site survey: adjust home positions to current drone positions

volfpeter:
@vasarhelyi We'll need a good, short label for the button that triggers this action. Feel free to commit it into this PR or directly to `dev`. Thanks!

### DIFF
--- a/src/algorithms/matching.js
+++ b/src/algorithms/matching.js
@@ -14,7 +14,7 @@ import sortBy from 'lodash-es/sortBy';
  *
  * @param {number[][]} matrix the input matrix
  * @param {object} parameters additional parameters to forward to the algorithm
- * @param {string} parameters.algorithm the algorithm to run; `hungarian` or `greedy`
+ * @param {string} algorithm the algorithm to run; `hungarian` or `greedy`
  * @returns the assignment in a matrix with N rows and 2 columns, representing
  *          the row and column indices of the matrix cells that were chosen
  */
--- a/src/algorithms/matching.js
+++ b/src/algorithms/matching.js
@@ -13,8 +13,8 @@ import sortBy from 'lodash-es/sortBy';
  * yields more intutive results.
  *
  * @param {number[][]} matrix the input matrix
- * @param {string} algorithm the algorithm to run; `hungarian` or `greedy`
  * @param {object} parameters additional parameters to forward to the algorithm
+ * @param {string} parameters.algorithm the algorithm to run; `hungarian` or `greedy`
  * @returns the assignment in a matrix with N rows and 2 columns, representing
  *          the row and column indices of the matrix cells that were chosen
  */
@@ -73,7 +73,6 @@ export function greedyMatchingAlgorithm(matrix, parameters = {}) {
   let numberOfRowsLeft = numberOfRows;
   let numberOfColumnsLeft = numberOfColumns;
 
-  // eslint-disable-next-line no-unused-vars
   for (const [rowIndex, colIndex, _] of sortedPairs) {
     if (!isRowUsed[rowIndex] && !isColUsed[colIndex]) {
       result.push([rowIndex, colIndex]);
--- a/src/features/site-survey/actions.ts
+++ b/src/features/site-survey/actions.ts
@@ -11,7 +11,7 @@ import { getBase64ShowBlob } from '~/features/show/selectors';
 import { showError } from '~/features/snackbar/actions';
 import { getCurrentGPSPositionsOfActiveUAVs } from '~/features/uavs/selectors';
 import messageHub from '~/message-hub';
-import { isGPSPosition, type GPSPosition } from '~/model/geography';
+import { isGPSPositionValid, type GPSPosition } from '~/model/geography';
 import {
   GROSS_CONVEX_HULL_AREA_ID,
   NET_CONVEX_HULL_AREA_ID,
@@ -233,7 +233,7 @@ export const adjustHomePositionsToDronePositions =
     const dronePositions = getCurrentGPSPositionsOfActiveUAVs(getState());
     if (
       homePositions === undefined ||
-      dronePositions.some((val) => !isGPSPosition(val))
+      dronePositions.some((val) => !isGPSPositionValid(val))
     ) {
       return;
     }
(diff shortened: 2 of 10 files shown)


src/algorithms/matching.js
```diff
@@ -13,8 +13,8 @@ import sortBy from 'lodash-es/sortBy';
  * yields more intutive results.
  *
  * @param {number[][]} matrix the input matrix
- * @param {string} algorithm the algorithm to run; `hungarian` or `greedy`
  * @param {object} parameters additional parameters to forward to the algorithm
+ * @param {string} parameters.algorithm the algorithm to run; `hungarian` or `greedy`
```
ntamas: This is weird; did the linter complain about the existing docstring? Strictly speaking, `algorithm` is not part of `parameters` because `parameters` itself is just the "rest" argument of the options object.

src/features/site-survey/actions.ts
```diff
@@ -214,6 +220,49 @@ export const updateModifiedFeatures = (
...
+ */
+export const adjustHomePositionsToDronePositions =
+  (): AppThunk => async (dispatch, getState) => {
+    // Only outdoor shows are supported by the dialog.
+
+    const distanceFunction = haversineDistance;
+    const homePositions = getHomePositionsInWorldCoordinates(getState());
+    const dronePositions = getCurrentGPSPositionsOfActiveUAVs(getState());
+    if (homePositions === undefined || dronePositions.includes(undefined)) {
+      return;
```
ntamas, line 235: A snackbar message would be nice here to let the user know why nothing is happening after clicking on the button.

src/features/site-survey/actions.ts
```diff
@@ -214,6 +220,49 @@ export const updateModifiedFeatures = (
...
+    }
+
+    const distances = calculateDistanceMatrix(
+      homePositions,
+      dronePositions as GPSPosition[], // Validated in guard clause above.
+      {
+        distanceFunction,
+        // eslint-disable-next-line @typescript-eslint/ban-types
+        getter: (item: GPSPosition | null): LonLat =>
```
ntamas: Can we ever receive `null` here? We excluded `undefined` above - maybe we should do the same with `null`?

src/features/site-survey/selectors.ts
```diff
@@ -293,3 +294,12 @@ export const getVisibleLayersInOrder = createSelector(
...
   }
 );
+
+export const selectAdjustHomePositionsToDronePositionsEnabled = createSelector(
+  getHomePositionsInWorldCoordinates,
+  getCurrentGPSPositionsOfActiveUAVs,
+  (homePositions, dronePositions) =>
+    homePositions !== undefined &&
+    dronePositions.length > 0 &&
+    !dronePositions.includes(undefined)
```
ntamas: I think we should introduce a helper function named `isGPSPositionValid()` or something like that, and exclude `null`, `undefined` and also the so-called "null island" (i.e. lat = 0, lon = 0). The latter might briefly be sent by ArduPilot when the drone has acquired GPS time sync but not a proper GPS fix yet. Check if we already have a similar function anywhere.

In particular, in the UAV list view we show "No position" or something similar for drones where we consider the GPS position not to be valid. Check what the conditions are for showing this message instead of a valid GPS position, and use the same condition here. You can introduce the helper function in `~/utils/geography.ts` if needed. If the UAV list does not treat the null island specially, then no need to treat it separately here either.

ntamas (changes requested):

src/algorithms/matching.js
volfpeter: The linter complained where I called this function. I guess since the second argument has no name, it automatically interpreted it as `parameters` from the docstring? No idea actually, but this change fixed the issue.

src/features/site-survey/actions.ts
volfpeter: I think we technically can, but I'll recheck. Since there are several calls and results here that come from untyped JS code, in one or two places I just preferred to be "safe" with the typing.

src/features/site-survey/selectors.ts
volfpeter: I'll check that bit of code and create a type guard. I didn't want to add too much extra typing and small utilities, but given this comment, I may add a couple, because there are a couple of places where a type guard would be a nicer option than casting (at a small performance cost).

src/algorithms/matching.js
volfpeter: I just checked it. If I revert the change and try to pass an algorithm in `adjustHomePositionsToDronePositions`, then I get the following TypeScript error: `Argument of type '{ algorithm: string; }' is not assignable to parameter of type 'string'.ts(2345)`

~~I tried a few other documentation options, but  it seems I must use `parameters` and `parameters.x` to not get an error. I guess TS infers the second argument's name from what follows the spread operator.~~

Okay, I found a better fix: change the order of the arguments in the docstring :slightly_smiling_face: If the docstring for `parameters` is before `algorithm`, then TS also doesn't complain. I've updated the docs like this.

src/features/site-survey/actions.ts
volfpeter: I added an `isGPSPositionValid()` type guard that does `null`, `undefined`, and null island checks only, so we don't do unnecessary calculation with `isGPSPosition()`. I also added the null island check to `isGPSPosition()` for consistency. It'd be nice to get a confirmation on whether this is okay.

src/features/site-survey/selectors.ts
volfpeter: It seems we do check for null island in `uav.ts`, in the `UAV.position` getter (and return `undefined` in that case).

The selector now checks positions with `isGPSPositionValid` which fails for the null island, so I think we're good.

src/features/site-survey/actions.ts
volfpeter: The button is disabled if these checks fail (I don't think the user should ever be able to execute this action in that case), but I can still add a snackbar message. We could also show a snackbar unconditionally at the end, displaying how many home positions were updated. I'm not yet sure about these UX preferences in the project.

ntamas: If this branch is not reachable via normal user action, then a `console.warn(...)` should be enough, no need for a snackbar message. Up to you.

ntamas (approved):
Thanks a lot!

src/features/site-survey/actions.ts
volfpeter: I added the warning (committed directly to `dev`).

PR #71, opened 2025-05-28
MUI v5 upgrade

volfpeter:
**Description**

We'll need to thoroughly test the application. There are still some bugs, todos, and open questions. New bugs and todos should be added to the PR description.

@vasarhelyi There some upgrade steps that can be done incrementally in the future. I recommend creating tech debt tickets for those in the issue tracker. The same may apply to some of the TODOs and questions. We'll see the feedback from @ntamas and @isti115 

Notes:

- It is assumed that the base theme and style utilities are fine, in `@skybrush/app-theme-mui` and `@skybrush/mui-components`. 
- Import groups were inconsistent in the codebase. I streamlined those in all changed files like this: i) 3rd party imports ii) "known" 3rd party imports (`@collmot`, `@skybrush`) iii) absolute imports iv) relative imports. I'd even consider merging groups iii and iv to match the Python standard and simplify things (imports would still be sorted, auto-import would have a better chance of being in the correct place).

Questions:

- What to do with the `<Tabs />` `indicatorColor` and `textColor` change? Both `primary` now. See the [upgrade docs](https://mui.com/material-ui/migration/v5-component-changes/#update-default-indicatorcolor-and-textcolor-prop-values)
- `FormControl` and `Select`: `variant='standard'` should be added to keep the old style, default style changed to `primary`. There's a codemod for this, ran it than went back to move closer to Material Design.
- How about `TextField`? We use `filled` in many places, and (explitict) `standard` and default in some others. For now everything remains as it was, except the new default is `outlined`.
- Consider whether we want to keep the custom `IconButton` `size='large'` prop? Default value changed, that's why these were set throughout the codebase.
- In some cases the default color changed from `secondary` to `primary` to better align the components with [Material Design guidelines](https://mui.com/material-ui/migration/v5-component-changes/#set-to-quot-primary-quot-by-default). In terms of maintenance, it's best to be as close to the "default" design as possible, but if we want to keep mimicking v4, then these components need to be updated.
- `FadeAndSlide.js`: is there a better way to upgrade `defaultTimeout` constants?

Bugs:

- [x] Styles break when opening a dialog (all caused by an outdated version of `mui-rff`):
    - [x] Manual tab in the Connection dialog
    - [x] Geofence dialog.
    - [x] Editing a location in the Locations panel.
    - [x] RTK dialog.
    - [x] Safety dialog.
    - [x] Background color issue in Toolbox / Offline maps dialog.
- [x] Site survey dialog: the interaction hint is too tall, it messes up the dialog's layout. The selected tab marker is not centered.
- [x] The "No show file loaded" text in the Control panel has a larger font size than other primary list item texts.
- [x] Clicking on the "Use secure connection" switch in the Connection dialog does nothing but the switch's style flickers. It color also seems wrong. (Also caused by outdated `mui-rff`)
- [x] Toolbox / Coordinate averaging dialog: after clicking on the + icon to add a drone, the popover does not respond to click events, but it selects one of the drones after Enter is pressed. The bug is in `SemanticAvatar` in `@skybrush/mui-components`, PR with the fix is ready, *this bug will be fixed when the PR is merged and a new version is released.*
- [x] Preferences dialog, Preflight and API keys tabs: textbox colors are off
- [x] Preferences dialog, bottom padding may be too large. *Fixed by removing the unnecessary padding and margins* from the individual tabs, but the correct solution would be remove all margins and paddings and use `gap` on the dialog content instead. 
- [x] Wrong textbox color in the Layers panel when editing the name of a layer.
- [x] UAV details panel, the UAV selector button in the top left corner seems unresponsive to clicks. *Same issue as in the coordinate averaging dialog.*
- [x] The font size of header widgets seem to have grown, which makes the header taller as well (may be related to the Site survey interaction hint size change). (Height is correct when the Preferences button is a rectangle). Will be fixed by [this PR](https://github.com/skybrush-io/skybrush.js/pull/6) in `skybrush.js`, if accepted.
- [x] Bottom padding is missing from text boxes in the "Edit saved location" dialog box (click on the pencil icon next to a location in the Locations panel). *Not a padding, but a layout and gap issue. Fixed, but not in the optimal way, that would need `grid` layout and additional changes. It should be good enough for now.*
- [x] _"Choose start time"_ button on the _"Show control"_ panel doesn't open the dialog
- [x] Letter spacing too large (extra letter spacing is removed from the app theme, waiting for new `@skybrush/app-theme-ui` release). 
- [x] `StartTimeDialog`: date and time inputs are not aligned vertically with the Reference dropdown. Also, all inputs should have `variant='filled'`. *This is an unfixed bug in `mui-rff`, here's my [fork with the fix](https://github.com/volfpeter/mui-rff/tree/fix/mui-v5-date-time-pickers-TextField-customization). Currently the issue is fixed by patching the lib. The fix is not compatible with `mui-rff` v6 because of the capital starting letter in the `TextFieldProps` in our fix, but lower-case t in the official, v6 fix. But it should be compatible with mui v7. As a result, there's no bugfix PR.*
- [x] Unnecessary paddings in Geofence dialog.
- [x] ~~Show/Hide empty mission slots in the UAVs panel's header has no tooltip.~~ There is a tooltip, but it's only visible when the button is enabled (same as before the upgrade).
- [x] RTK dialog (requires server connection) layout is broken.
- [x] `DialogHeaderListItem` `backgroundColor` is not correct, see for example the Toolbox/Offline Maps dialog.
- [ ] The disabled state of the Show/Hide empty mission slots button (when "sort by mission ID" is off) is not correct. It seems the selected style overrides the disabled style. Dumb fix would be to make the button visibly deselected when it is disabled, but then we'd lose some visual feedback (disabled&selected should be slightly different to plain disabled).
- ~~[ ] From Gábor (Win10+Firefox): the right edge of the UAVs panel is off, it should be aligned with the X icon in the top right. Looks good to me (Fedora+Firefox).~~
- [x] UAVs toolbar seems to be a bit more dense than previously.
- [x] Other dialogs may also have padding issues. (It seems `DialogContent` got some vertical padding, but it's not mentioned in the docs.) *For now this is fixed by using a new `CompactDialogContent` component in most cases. With my limited experience with the codebase, I'm reluctant to blindly replace every `DialogContent`, as in some cases the extra padding is necessary. For similar reason's I haven't touched the theme itself.*
- [x] Icons are missing from the date and time inputs of Start time dialog. 
- [x] Upgrade `material-ui-popup-state`.

TODOs:

- [x] Check what to do with `disableGutters` props.
- [ ] Test the `AutoComplete` component to make sure `inputRef` works correctly. See the [upgrade docs](https://mui.com/material-ui/migration/v5-component-changes/#forward-ref-instead-of-inputref-prop).
- [x] After clicking on the UAVs by status widget in the header ends with a warning "UAVs panel not added to the workbench yet", but nothing should be displayed here. Clicking on a line in the widget's tooltip works correctly: corresponding drones are selected. It may be unrelated to the MUI upgrade.
- [ ] When double-clicking a drone on the map for the first time `Warning: React does not recognize the `DraggableProps` prop on a DOM element.`. It may not be a new issue.

For @ntamas :

- [x] Seems fixed, but pls check it, maybe it behaves differently on your machine/browser: UAV details dialog: when first opened (double click on drone), the UI behind the dialog shifts down by about 200px. Closing and reopening fixes the issue (maybe only for the same drone).

Q/A, Testing:

- [ ] `UAVSelector`: arrow position when the popover is pushed against the edge of the viewport
- [ ] Event log component (complex virtualized list)


volfpeter:
> What's the status of this one? Is this ready for a round of reviews?

I think so, yes. The status is basically what you see in the description. Unfortunately I had no time last week to work on those. There shouldn't be any serious issues, but I'll also review everything when I'm back from holiday.

volfpeter:
@ntamas @vasarhelyi All known issues seem to be fixed now. I hope this is the last time I ask for review and QA :slightly_smiling_face:

PR #76, opened 2025-06-26
Range selection in PendingUAVIdOverlay

volfpeter:
Notes:

- The range separator is `:`.
- For range selection, both the start and end index must be set, either in `1:1` or `s11:22` format. Shorthands like `10:`, `:20`, or `:` (for select all) are not supported right now.

The redux actions seem to be missing for adding IDs to and removing IDs from the selection (with e.g. `-` and `+` prefixes), and this feature is not high priority, so I didn't implement it. If needed, I can add it to the PR.


ntamas:
Re the `-` vs `:` debate: @vasarhelyi's comment about the fact that `-` is usually accessible without a modifier is relevant here: sometimes it happens in field tests that the operator does not have both hands available for typing on the keyboard. It is not ideal, but it happens; imagine holding the GCS laptop in one of your hands while trying to type with the other. So, not having a modifier key attached to the range selection is advantageous.

I can easily let go of the idea of using `+` or `-` as a prefix if it complicates things.

volfpeter:
Okay, the separator will be `-` then. I'll update the PR next week.

PR #77, opened 2025-06-27
Drone selection in takeoff area setup dialog

volfpeter:
@vasarhelyi Let me know what you think about the UI (I didn't want to make the entire row clickable, it's a bit hidden and easy to accidentally press). Selection is now supported in the Missing, Misplaced, and Misaligned lists, but I'm not sure if it makes sense in the Missing list. Please let me know if I should remove the feature from that list.

### DIFF
--- a/src/views/show-control/TakeoffAreaSetupDialog.jsx
+++ b/src/views/show-control/TakeoffAreaSetupDialog.jsx
@@ -1,15 +1,15 @@
-import PropTypes from 'prop-types';
-import React from 'react';
-import { connect } from 'react-redux';
-import { Translation, withTranslation } from 'react-i18next';
-
 import Box from '@material-ui/core/Box';
 import Button from '@material-ui/core/Button';
 import DialogActions from '@material-ui/core/DialogActions';
 import DialogContent from '@material-ui/core/DialogContent';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import Switch from '@material-ui/core/Switch';
+import { SelectAll } from '@material-ui/icons';
 import VerticalAlignBottom from '@material-ui/icons/VerticalAlignBottom';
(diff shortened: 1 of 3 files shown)


ntamas:
Hmmm I think it doesn't really make sense in the "Missing" list - sure, it would select the empty slots in the UAVs list but there's really not much you can do with them. I would leave that out.

src/views/show-control/TakeoffAreaSetupDialog.jsx
```diff
@@ -1,15 +1,15 @@
...
-import { connect } from 'react-redux';
-import { Translation, withTranslation } from 'react-i18next';
-
 import Box from '@material-ui/core/Box';
 import Button from '@material-ui/core/Button';
 import DialogActions from '@material-ui/core/DialogActions';
 import DialogContent from '@material-ui/core/DialogContent';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import Switch from '@material-ui/core/Switch';
+import { SelectAll } from '@material-ui/icons';
```
ntamas: We usually import icons directly from the corresponding file (i.e. `@material-ui/icons/SelectAll` in this case). To be honest I don't know whether it's necessary - the reason why we do it this way is to avoid bundling icons from `@material-ui/icons` that we don't need. It used to help in the past when tree-shaking routines were not as smart as they are today.

ntamas (changes requested):
Only one tiny change - feel free to merge afterwards.

src/views/show-control/TakeoffAreaSetupDialog.jsx
volfpeter: Ah, auto-import. Sorry, I didn't notice it! Fixed.

PR #86, opened 2025-07-09
feat: optionally restrict UploadDialog to global selection

volfpeter:
@ntamas @isti115 During the review, please pay extra attention to how selectors and actions changes. There may be some use-cases I'm not aware of (thinking of job scopes, compatible drones, etc.).

@vasarhelyi Let me know what you think about the UI.

Short summary of how the feature works:

- There's a switch component now in the top right corner of the Upload dialog with a text next to it that describes its role.
- If the switch is checked, only the drones in the application's global selection will be visible in the dialog. The checked state of this button is preserved.
- If the user haven't selected drones manually (= put them to waiting state), then the upload job will only be triggered for the visible drones. Otherwise it will be triggered for the waiting drones, disregarding whether they are visible or not.
- Row headers now show the formatted IDs of the drones in the given row (previously this only worked because every drone was displayed).

The global selection hotkeys (notably range selection) works in the dialog, so a convenient user flow is to restrict the dialog to the global selection and then 1) use range selection hotkeys to change what's visible 2) click Start 3) goto step 1 or close the dialog.

### DIFF
--- a/src/features/upload/UploadStatusLights.jsx
+++ b/src/features/upload/UploadStatusLights.jsx
@@ -10,15 +10,14 @@ import { makeStyles } from '@material-ui/core/styles';
 
 import BackgroundHint from '@skybrush/mui-components/lib/BackgroundHint';
 
-import { getMissionMapping } from '~/features/mission/selectors';
-import { getUAVIdList } from '~/features/uavs/selectors';
-import { formatMissionIdRange } from '~/utils/formatting';
-
 import { toggleUavsInWaitingQueue } from './actions';
 import { JobScope } from './jobs';
 import {
+  getMissionIdFormatter,
+  getMissionMapping,
   getObjectIdsCompatibleWithSelectedJobInUploadDialog,
   getScopeOfSelectedJobInUploadDialog,
+  getUAVIdList,
 } from './selectors';
 import UploadStatusPill from './UploadStatusPill';
 import UploadStatusRowHeader from './UploadStatusRowHeader';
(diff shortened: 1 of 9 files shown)


vasarhelyi:
I think it would be more informative to have text next to the toggle switch at the top right corner corresponding to the different states, such as in e.g. the light control panel. Some alternatives:

1
all drones [toggle on/off] global selection

2.
All drones. Click to see global selection [off toggle]
Global selection. Click to see all drones [on toggle]

And the title could be changed as well, corresponding to the toggle state somehow.

src/features/upload/UploadStatusLights.jsx
```diff
@@ -50,17 +49,21 @@ const useStyles = makeStyles((theme) => ({
...
 /**
- * Default formatter used to create the row headings when the user specified
- * no formatter.
+ * Returns the row header text for the given items.
+ *
+ * @param {Array} items  The array of items in the row.
+ * @param {function} idFormatter  Function that formats IDs for the row header.
  */
-const defaultRowHeaderFormatter = (_start, _end, items) => {
+const rowHeaderForItems = (items, idFormatter = formatId) => {
```
ntamas: Consider moving this back to `utils/formatting.ts` under a name like `formatIdRange()` just in case we need similar functionality elsewhere. This function looks generic enough that it ideserves its own place in `utils/formatting.ts` (which is the first place I look for a generic formatter function) instead of being tucked away in `src/features/upload/UploadStatusLights.jsx`.

volfpeter:
Thanks for the review. I still need to finish the tasks from my previous comment. This will be the first item on my todo list when I'm back.

src/features/upload/UploadStatusLights.jsx
volfpeter: Done (not too happy with the name though).

volfpeter:
@ntamas @vasarhelyi Pls have a look at this and let me know if the current solution is okay.

https://github.com/user-attachments/assets/4c6b7f14-b6d1-4428-ad90-e97369f69d8e

vasarhelyi:
Looks great for me, thanks!

PR #95, opened 2025-08-22
Feat/show adapt takeoff duration

vasarhelyi:
This PR adds takeoff duration parameter to the show adapt dialog to have the option to specify explicit length for the takeoff and thus for the show start. TODOs:

- [ ] Would be nice to show the original takeoff/show/rth segment durations somewhere in the dialog for reference (if someone wants to keep the exact takeoff length, it needs two iterations now: first one shows diff, second corrected can result in diff=0)
- [ ] is the "0 means undefined" rule in the general number input box OK for takeoff duration? It was simpler to include the duration in a `createNumericField()` call than implementing a new `createOptionalNumericField()` ...

### DIFF
--- a/src/features/show-configurator/AdaptParametersForm.tsx
+++ b/src/features/show-configurator/AdaptParametersForm.tsx
@@ -9,6 +9,7 @@ import FormHeader from '@skybrush/mui-components/lib/FormHeader';
 
 import {
   SimpleDistanceField,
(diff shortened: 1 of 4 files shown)


volfpeter (approved):
Looks good. We could create an issue for the TODO.

vasarhelyi:
(I created the issue for the first todo item)

PR #98, opened 2025-09-02
MissionInfoLayer: use min distance selectors from the show state when the app is in show context

volfpeter:
This way, in show context, we only do 2 quick calculations instead of 2 quick and 2 long ones when a show is loaded or transformed.

The reason for putting the extra selectors to `mission-info.jsx` is that:

- some selectors in the `show` slice already depend on selectors in the `mission` slice, so adding these selectors to the `mission` slice (where they would normally belong) would create a circular dependency
- the existing min distance selectors were only used in `mission-info.jsx`.

Alternatively, we could split `mission/selectors.ts` into two files to break the circular dependency. The benefit would be that we could hide the optimization logic inside the exported `getMinimumDistance*` selectors, and then the optimization would automatically work wherever these selectors are used in the future.

Let me know what you think.

*Note: I still never tested the app in mission mode. It wouldn't hurt in this case, but this change should have no effect for that use-case.*

### DIFF
--- a/src/views/map/layers/mission-info.jsx
+++ b/src/views/map/layers/mission-info.jsx
@@ -1,6 +1,7 @@
 import Checkbox from '@mui/material/Checkbox';
 import FormControlLabel from '@mui/material/FormControlLabel';
 import FormGroup from '@mui/material/FormGroup';
+import { createSelector } from '@reduxjs/toolkit';
 import dropWhile from 'lodash-es/dropWhile';
 import takeWhile from 'lodash-es/takeWhile';
 import unary from 'lodash-es/unary';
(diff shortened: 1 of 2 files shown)


src/views/map/layers/mission-info.jsx
```diff
@@ -78,6 +82,36 @@ import {
...
+const getMinimumHomePositionDistanceSelector = createSelector(
+  getMissionType,
+  (type) =>
+    type === MissionType.SHOW
+      ? getMinimumDistanceBetweenTakeoffPositionsInShow
+      : getMinimumDistanceBetweenHomePositionsInMission
+);
+
+const getMinimumDistanceBetweenHomePositions = createSelector(
+  (state) => state,
```
ntamas, line 96: This selector would become invalid when _any_ property changes _anywhere_ in the state, even if that property is completely unrelated to the minimum distance between home positions. This kinda defeats the whole point of using `createSelector(...)`.

Give me a day or so to decide whether I can come up with a better solution or not.

ntamas:
Decided to use an alternative implementation in 7ee8dd7b. Thanks for your work on this!

src/views/map/layers/mission-info.jsx
volfpeter: That's right, but all this selector does is execute another selector which is memoized, so the overhead is minimal.

PR #102, opened 2025-09-20
Remove the tour feature

volfpeter:
To let us remove the `reactour` dependency.

I did a text search on the entire codebase, I hope it was enough.


config/webapp-demo.tsx
ntamas, line 1: Do not remove the `webapp-demo` config file, only the tour should be removed.

src/components/header/HelpButton.jsx
```diff
@@ -15,7 +15,7 @@ const HelpButton = () => (
   <Translation>
     {(t) => (
       <GenericHeaderButton
-        id='tour-help-button'
+        id='help-button'
```
ntamas: This ID was required for the tour only so do not rename it - remove the entire ID instead.

assets/css/screen.less
ntamas, line 1: The file was also reformatted accidentally as a side effect but I do not mind.

src/views/light-control/LightControlPanel.jsx
```diff
@@ -12,7 +12,7 @@ import LightControlMainSwitch from './LightControlMainSwitch';
  */
 const LightControlPanel = () => (
   <Box
-    id='tour-light-control'
+    id='light-control-panel'
```
ntamas: Same here - the ID was needed for the tour only so it acn probably be removed in its entirety.

src/views/show-control/ShowControlPanel.jsx
```diff
@@ -19,7 +19,7 @@ import TakeoffAreaSetupDialog from './TakeoffAreaSetupDialog';
  */
 const ShowControlPanel = () => (
   <Box
-    id='tour-show-control'
+    id='show-control-panel'
```
ntamas: Same here - the ID was needed for the tour only so it acn probably be removed in its entirety.

ntamas (changes requested):

assets/css/screen.less
volfpeter: Yeah, I can revert that if you prefer. It was just prettier with format on save.

config/webapp-demo.tsx
volfpeter: Ah, okay. I restored it. It seemed like it's only a tour config, but it still has a bit of content without the tour.

src/components/header/HelpButton.jsx
volfpeter: Yes, I guessed. The reason I kept these IDs is they may be helpful for example for testing or creating screenshots with puppeteer. But we can add them if we need them. I'll remove them.

*Edit: done.*

ntamas (approved):

PR #104, opened 2025-10-29
feat(rtk): position saving and reloading

zarcell:
This still needs testing on field, use this only for code review!


src/features/rtk/utils.ts
```diff
@@ -88,7 +90,7 @@ const descriptions: Record<string, string> = {
   'rtcm3/1121': 'BeiDou MSM1 (DGNSS pseudorange)',
   'rtcm3/1122': 'BeiDou MSM2 (RTK pseudorange)',
   'rtcm3/1123': 'BeiDou MSM3 (code, carrier)',
-  'rtcm3/1124': 'BeiDou MSM4 (code, carrier, CNR)',
+  'rtcm3/1124': 'BeiDou MSM3 (code, carrier, CNR)',
```
isti115: Do I assume correctly that this was modified accidentally, or is it an intentional change? 🤔

src/features/rtk/AntennaPositionIndicator.jsx
```diff
@@ -46,15 +50,32 @@ const AntennaPositionIndicator = ({
           </IconButton>
         </Tooltip>
       )}
-    </>
+      {onShowSavedCoordinates && (
+        <Tooltip
+          content={t('antennaPositionIndicator.useSavedCoordinate')}
+          trigger='mouseenter'
```
isti115, line 58: Is this actually necessary? Based on the [Tippy source](https://github.com/atomiks/tippyjs/blob/v6.3.7/src/props.ts#L63) the default seems to be `'mouseenter focus'`, what do we gain by dropping `focus`? I would even think that this seems to go against touchscreen support.

src/features/rtk/AntennaPositionIndicator.jsx
```diff
@@ -46,15 +50,32 @@ const AntennaPositionIndicator = ({
...
+        >
+          <IconButton
+            disabled={!hasSavedCoordinates}
+            size='large'
+            sx={{ ml: -1.5 }}
```
isti115: This seems like a hacky way to fix the placement, proper spacing could probably be achieved in a cleaner way, but I guess that it's fine for the time being.

src/features/rtk/RTKCoordinateRestorationDialog.jsx
```diff
@@ -0,0 +1,123 @@
...
+
+import { closeCoordinateRestorationDialog } from '~/features/rtk/slice';
+import {
+  getCoordinateRestorationDialog,
+  getFormattedCoordinatePosition,
+  getSavedCoordinatesForPreset,
+} from '~/features/rtk/selectors';
+import { useSavedCoordinateForPreset } from '~/features/rtk/actions';
+
+const SavedCoordinateItem = connect((state, { coordinate }) => ({
```
isti115: It's more conventional to separately define the "dumb" component first and then connect it to the redux store afterwards. In his specific case I personally don't even think that `SavedCoordinateItem` should be connected to the store at all, as it will be rendered multiple times and it seems unnecessary to create a separate coordinate formatter for each of them. I would recommend following the pattern from [`MeasurementList`](https://github.com/skybrush-io/live/blob/ca0c079c572255337be155340b2e7e151fbcc3c9/src/features/measurement/MeasurementList.jsx) instead, where the list component gets a single formatter and passes it forward to the list items.

src/features/rtk/RTKCoordinateRestorationDialog.jsx
```diff
@@ -0,0 +1,123 @@
...
+}))(({ coordinate, formattedPosition, onClick }) => {
+  const { accuracy, savedAt } = coordinate;
+  const savedDateTime = formatDate(new Date(savedAt), 'yyyy-MM-dd HH:mm:ss');
+
+  return (
+    <ListItem disablePadding>
+      <ListItemButton onClick={() => onClick(coordinate)}>
+        <ListItemText
+          primary={formattedPosition}
+          secondary={`Accuracy: ${accuracy.toFixed(3)} m • ${savedDateTime}`}
```
isti115: We have a dedicated [`formatDistance`](https://github.com/skybrush-io/live/blob/ca0c079c572255337be155340b2e7e151fbcc3c9/src/utils/formatting.ts#L230) function (in `src/utils/formatting.ts`) that will e.g. automatically switch to centimeters based on the accuracy.

src/features/rtk/RTKCoordinateRestorationDialog.jsx
```diff
@@ -0,0 +1,123 @@
...
+    </ListItem>
+  );
+});
+
+const RTKCoordinateRestorationDialog = ({
+  t,
+  dialog,
+  savedCoordinates,
+  onClose,
+  onUseSaved,
```
isti115: Sort these alphabetically, at least the part before the event handlers. Personally I think that those shouldn't be treated specially either (they will be naturally grouped due to their "on" prefix anyway), but I'm not sure whether there's a consensus regarding this in the team.

```suggestion
  dialog,
  onClose,
  onUseSaved,
  savedCoordinates,
  t,
```

_(Edit: A quick look through the results of `rg -U '(?s)^const \S* = \(\{.*?\}'` seem to confirm the the event-listeners-at-the-end approach has mostly been abandoned by now.)_

src/features/rtk/RTKCoordinateRestorationDialog.jsx
```diff
@@ -0,0 +1,123 @@
...
+}) => {
+  if (!dialog.open) {
+    return null;
+  }
+
+  const { presetId } = dialog;
+
+  const handleUseSaved = (coordinate) => {
+    onClose(); // Close dialog first
+    onUseSaved(presetId, coordinate); // Then start async operation
```
isti115: The linter is currently ignoring all JavaScript files (which is a bit questionable IMO, since currently only about half or our codebase is TS), but it would most likely complain about an unused promise here.

src/features/rtk/RTKCoordinateRestorationDialog.jsx
```diff
@@ -0,0 +1,123 @@
...
+const RTKCoordinateRestorationDialog = ({
+  t,
+  dialog,
+  savedCoordinates,
+  onClose,
+  onUseSaved,
+}) => {
+  if (!dialog.open) {
+    return null;
+  }
```
isti115: Isn't it enough to control the visibility of the dialog with `open={dialog.open}`?

src/features/rtk/RTKCoordinateRestorationDialog.jsx
```diff
@@ -0,0 +1,123 @@
...
+    >
+      <DialogTitle id='coordinate-restoration-dialog-title'>
+        {t('RTKCoordinateRestorationDialog.title')}
+      </DialogTitle>
+      <DialogContent>
+        <Box>
+          {savedCoordinates.length === 0 ? (
+            <Box p={2} textAlign='center'>
+              {t('RTKCoordinateRestorationDialog.noSavedCoordinates')}
+            </Box>
```
isti115: There's a `BackgroundHint` component in `@skybrush/mui-components` for this purpose.

src/features/rtk/RTKCoordinateRestorationDialog.jsx
```diff
@@ -0,0 +1,123 @@
...
+                  onClick={handleUseSaved}
+                />
+              ))}
+            </List>
+          )}
+        </Box>
+      </DialogContent>
+      <DialogActions>
+        <Button color='primary' onClick={onClose}>
+          {t('RTKCoordinateRestorationDialog.cancel')}
```
isti115: Don't duplicate common strings, reuse the one defined under `general`!
```suggestion
          {t('general.action.cancel')}
```

src/i18n/en.json
```diff
@@ -844,4 +854,4 @@
     "timezone": "Timezone",
     "utcOffset": "UTC offset"
   }
-}
+}
```
isti115: This file is missing a newline character at the end, which makes me think that formatting in your editor is not set up properly. (I'll make sure to add a CI workflow for `prettier` too!)

src/i18n/en.json
```diff
@@ -21,6 +21,15 @@
     "passed": "Onboard preflight checks passed.",
     "signOffOn": "Sign off on onboard preflight checks"
   },
+  "RTKCoordinateRestorationDialog": {
+    "accuracyLabel": "Accuracy",
+    "cancel": "Cancel",
```
isti115: This is unnecessary, see my previous comment on `RTKCoordinateRestorationDialog.jsx`.

src/features/rtk/RTKCoordinateRestorationDialog.jsx
```diff
@@ -0,0 +1,123 @@
...
+    </Dialog>
+  );
+};
+
+RTKCoordinateRestorationDialog.propTypes = {
+  t: PropTypes.func,
+  dialog: PropTypes.object,
+  savedCoordinates: PropTypes.array,
+  onClose: PropTypes.func,
+  onUseSaved: PropTypes.func,
```
isti115: Sort, as above:
```suggestion
  dialog: PropTypes.object,
  onClose: PropTypes.func,
  onUseSaved: PropTypes.func,
  savedCoordinates: PropTypes.array,
  t: PropTypes.func,
```

src/features/rtk/RTKCorrectionSourceSelector.jsx
```diff
@@ -19,7 +22,11 @@ const nullPreset = {
   title: 'RTK disabled',
 };
 
-const RTKCorrectionSourceSelector = ({ onSourceChanged, t }) => {
+const RTKCorrectionSourceSelector = ({
+  onSourceChanged,
+  t,
+  setCurrentRTKPresetId,
```
isti115, line 27: Sort alphabetically.
```suggestion
  onSourceChanged,
  setCurrentRTKPresetId,
  t,
```

src/features/rtk/RTKCorrectionSourceSelector.jsx
```diff
@@ -31,29 +38,48 @@ const RTKCorrectionSourceSelector = ({ onSourceChanged, t }) => {
...
   const loading = presetsState.loading || selectionState.loading;
-  const hasError = presetsState.error || selectionState.error;
+  const hasError = presetsState.error;
 
-  const presets = presetsState.value ? presetsState.value : [];
+  if (selectionState.error) {
+    console.warn('Failed to load RTK selection state:', selectionState.error);
+  }
+
+  const presets = presetsState.value || [];
```
isti115: Prefer nullish coalescing.
```suggestion
  const presets = presetsState.value ?? [];
```

src/features/rtk/RTKCorrectionSourceSelector.jsx
```diff
@@ -31,29 +38,48 @@ const RTKCorrectionSourceSelector = ({ onSourceChanged, t }) => {
...
   const hasSelectionFromServer = selectionState.value !== undefined;
   const selectedOnServer =
-    selectionState.value !== undefined
-      ? selectionState.value === null
-        ? NULL_ID
-        : selectionState.value
-      : undefined;
+    selectionState.value === undefined
+      ? undefined
+      : (selectionState.value ?? NULL_ID);
```
isti115: I would rather keep the original logic here:
```suggestion
    selectionState.value !== undefined
      ? selectionState.value === null
        ? NULL_ID
        : selectionState.value
      : undefined;
```
It's a bit more verbose, but makes it clear that `null` is handled specially. (It took me non-zero cognitive effort to figure out why the sorter version was even necessary.)

src/features/rtk/RTKCorrectionSourceSelector.jsx
```diff
@@ -31,29 +38,48 @@ const RTKCorrectionSourceSelector = ({ onSourceChanged, t }) => {
...
 
     if (onSourceChanged) {
       onSourceChanged();
     }
   };
 
+  // Update current preset ID in Redux
+  useEffect(() => {
+    const currentId =
+      selectedByUser || (hasSelectionFromServer ? selectedOnServer : undefined);
```
isti115: This is equivalent to `selectedByUser ?? selectedOnServer`, right? 🤔

src/features/rtk/RTKCorrectionSourceSelector.jsx
```diff
@@ -31,29 +38,48 @@ const RTKCorrectionSourceSelector = ({ onSourceChanged, t }) => {
...
+    // Convert NULL_ID to undefined or keep as is? The selector uses 'undefined' for no selection usually.
+    // But presets have IDs.
+    const effectiveId = currentId === NULL_ID ? undefined : currentId;
+    setCurrentRTKPresetId(effectiveId);
+  }, [
+    selectedByUser,
+    selectedOnServer,
+    hasSelectionFromServer,
+    setCurrentRTKPresetId,
```
isti115: Sort!
```suggestion
    hasSelectionFromServer,
    selectedByUser,
    selectedOnServer,
    setCurrentRTKPresetId,
```

src/features/rtk/RTKCorrectionSourceSelector.jsx
```diff
@@ -144,13 +170,14 @@ const RTKCorrectionSourceSelector = ({ onSourceChanged, t }) => {
 RTKCorrectionSourceSelector.propTypes = {
   onSourceChanged: PropTypes.func,
   t: PropTypes.func,
+  setCurrentRTKPresetId: PropTypes.func,
```
isti115, line 168: Sort!
```suggestion
  onSourceChanged: PropTypes.func,
  setCurrentRTKPresetId: PropTypes.func,
  t: PropTypes.func,
```

src/features/rtk/RTKSatelliteObservations.jsx
```diff
@@ -175,7 +175,7 @@ const RTKSatelliteObservations = ({ height = 160, items }) => {
         items ? createDataFromItemsAndDrawingContext(items, chart.ctx) : NO_DATA
       );
     }
-  }, [chartRef.current, items]);
+  }, [items]);
```
isti115: Why did `chartRef.current` get removed? 🤨
It seems like it was [added on purpose](https://github.com/skybrush-io/live/commit/7388a2093f5f67eeea34594226da57424525699f#diff-bba0243e0c635dd7795857f300d3deb3a421d066a9a085487af0f7f08e0fddfeR178) by @ntamas in 7388a20.
Did something change or was it unnecessary in the first place?

src/store/index.js
```diff
@@ -113,6 +112,9 @@ const persistConfig = {
     // Store only the persistent settings of the upload procedure
     createFilter('upload', ['settings']),
 
+    // Store only saved coordinates from RTK
+    createFilter('rtk', ['savedCoordinates']),
```
isti115, line 116: Shouldn't we maybe store the selected preset ID and the antenna position format preference of the user as well? Or is it important for some reason to explicitly selects the preset each time? (Would it for example not be synchronized to the server when loaded from the saved state?)

src/features/rtk/RTKSetupDialogBottomPanel.jsx
```diff
@@ -15,8 +15,16 @@ import { Tooltip } from '@skybrush/mui-components';
 
 import FadeAndSlide from '~/components/transitions/FadeAndSlide';
 
-import { getSurveyStatus, shouldShowSurveySettings } from './selectors';
-import { toggleSurveySettingsPanel } from './slice';
+import {
+  getSurveyStatus,
+  shouldShowSurveySettings,
+  hasSavedCoordinateForPreset,
+  getCurrentRTKPresetId,
```
isti115: Sort alphabetically.
```suggestion
  getCurrentRTKPresetId,
  getSurveyStatus,
  hasSavedCoordinateForPreset,
  shouldShowSurveySettings,
```

src/features/rtk/RTKSetupDialogBottomPanel.jsx
```diff
@@ -15,8 +15,16 @@ import { Tooltip } from '@skybrush/mui-components';
...
+} from './selectors';
+import {
+  toggleSurveySettingsPanel,
+  showCoordinateRestorationDialog,
```
isti115: Sort alphabetically.
```suggestion
  showCoordinateRestorationDialog,
  toggleSurveySettingsPanel,
```

src/features/rtk/RTKSetupDialogBottomPanel.jsx
```diff
@@ -64,6 +75,12 @@ const RTKSetupDialogBottomPanel = ({
     }
   }, [onToggleSurveySettings, surveySettingsVisible, surveyStatus?.supported]);
 
+  const handleShowSavedCoordinates = () => {
+    if (currentPresetId) {
+      onShowSavedCoordinates(currentPresetId);
```
isti115: If `onShowSavedCoordinates` is not guaranteed to be a function (I don't really see a way it could end up being anything else, but based on the check below I guess it somehow could?) it would be safer to use optional chaining on the call.
```suggestion
      onShowSavedCoordinates?.(currentPresetId);
```

src/features/rtk/RTKSetupDialogBottomPanel.jsx
```diff
@@ -114,20 +136,31 @@ const RTKSetupDialogBottomPanel = ({
...
-    surveyStatus: getSurveyStatus(state),
-    surveySettingsVisible: shouldShowSurveySettings(state),
-  }),
+  (state) => {
+    const currentPresetId = getCurrentRTKPresetId(state);
+    return {
+      currentPresetId,
+      hasSavedCoordinates: currentPresetId
+        ? hasSavedCoordinateForPreset(state, currentPresetId)
+        : false,
```
isti115: Code returning a boolean constant from one side of a ternary can always be simplified into a logical expression.
```suggestion
      hasSavedCoordinates:
        Boolean(currentPresetId) &&
        hasSavedCoordinateForPreset(state, currentPresetId),
```

src/features/rtk/RTKSetupDialogBottomPanel.jsx
```diff
@@ -95,7 +112,12 @@ const RTKSetupDialogBottomPanel = ({
             )}
             <SurveyStatusIndicator {...surveyStatus} />
             <Box sx={{ flex: '1' }} />
-            <AntennaPositionIndicator />
+            <AntennaPositionIndicator
+              hasSavedCoordinates={hasSavedCoordinates}
+              onShowSavedCoordinates={
+                onShowSavedCoordinates ? handleShowSavedCoordinates : undefined
```
isti115: This check becomes redundant if we have conditional chaining at the call site. (Looking at `AntennaPositionIndicator` it seems to me like this isn't necessary for downstream either.)

src/features/rtk/RTKStatusUpdater.jsx
```diff
@@ -3,10 +3,12 @@ import isNil from 'lodash-es/isNil';
 import mapValues from 'lodash-es/mapValues';
 import PropTypes from 'prop-types';
 import { useEffect } from 'react';
-import { connect } from 'react-redux';
+import { connect, useDispatch, useSelector } from 'react-redux';
 
 import handleError from '~/error-handling';
 import { updateRTKStatistics } from '~/features/rtk/slice';
+import { saveCurrentCoordinateForPreset } from '~/features/rtk/actions';
+import { hasValidFix, shouldSaveCoordinate } from '~/features/rtk/utils';
```
isti115, line 12: Sort.
```suggestion
import { saveCurrentCoordinateForPreset } from '~/features/rtk/actions';
import { updateRTKStatistics } from '~/features/rtk/slice';
import { hasValidFix, shouldSaveCoordinate } from '~/features/rtk/utils';
```

src/features/rtk/RTKStatusUpdater.jsx
```diff
@@ -15,19 +17,40 @@ import useMessageHub from '~/hooks/useMessageHub';
  */
 const RTKStatusUpdater = ({ onStatusChanged, period = 1000 }) => {
   const messageHub = useMessageHub();
+  const dispatch = useDispatch();
+  const savedCoordinates = useSelector((state) => state.rtk.savedCoordinates);
```
isti115: Creating a dedicated selector for this might end up being more flexible than hardcoding the state access path.

src/features/rtk/utils.ts
```diff
@@ -118,3 +120,50 @@ export function formatSurveyAccuracy(
...
+ *
+ * @param status - The RTK statistics object.
+ * @returns True if a valid fix is present, false otherwise.
+ */
+export function hasValidFix(status: Partial<RTKStatistics>): boolean {
+  const hasECEF = Array.isArray(status?.antenna?.positionECEF);
+  const accuracy = status?.survey?.accuracy;
+  const flags = status?.survey?.flags;
+  const surveyedCoordinateValid =
+    typeof flags === 'number' ? (flags & 0b100) !== 0 : false;
```
isti115: Code returning a boolean constant from one side of a ternary can always be simplified into a logical expression.
```suggestion
    typeof flags === 'number' && (flags & 0b100) !== 0;
```

src/features/rtk/utils.ts
```diff
@@ -118,3 +120,50 @@ export function formatSurveyAccuracy(
...
+ * Determines whether the current coordinate should be saved for a given preset.
+ *
+ * @param status - The RTK statistics object containing the current antenna position.
+ * @param savedCoordinates - The record of saved coordinates keyed by preset ID.
+ * @param presetId - The ID of the preset to check against.
+ * @returns True if the coordinate is new and should be saved, false otherwise.
+ */
+export function shouldSaveCoordinate(
+  status: Partial<RTKStatistics>,
+  savedCoordinates: Record<string, RTKSavedCoordinate[]> | undefined,
```
isti115: Why do we need to accept `undefined` here? 🤔

src/features/rtk/RTKStatusUpdater.jsx
```diff
@@ -43,7 +66,7 @@ const RTKStatusUpdater = ({ onStatusChanged, period = 1000 }) => {
       valueHolder.finished = true;
       valueHolder.promise = null;
     };
-  }, [messageHub, onStatusChanged, period]);
+  }, [messageHub, onStatusChanged, period, dispatch, savedCoordinates]);
```
isti115: Sort!
```suggestion
  }, [dispatch, messageHub, onStatusChanged, period, savedCoordinates]);
```

src/features/rtk/actions.js
```diff
@@ -7,7 +7,11 @@ import {
   getFormattedAntennaPosition,
   isShowingAntennaPositionInECEF,
 } from './selectors';
-import { closeSurveySettingsPanel, setAntennaPositionFormat } from './slice';
+import {
+  closeSurveySettingsPanel,
+  setAntennaPositionFormat,
+  saveCoordinateForPreset,
```
isti115, line 13: Sort.
```suggestion
  closeSurveySettingsPanel,
  saveCoordinateForPreset,
  setAntennaPositionFormat,
```

src/features/rtk/actions.js
```diff
@@ -37,3 +41,102 @@ export const toggleAntennaPositionFormat = () => (dispatch, getState) => {
...
+      await messageHub.execute.setRTKAntennaPosition({
+        position: savedCoordinate.positionECEF,
+        accuracy: savedCoordinate.accuracy,
+      });
+
+      // Check if the component is still mounted by checking if the dialog is still open
+      const currentState = getState();
+      if (!currentState.rtk.dialog.coordinateRestorationDialog.open) {
+        return;
+      }
```
isti115: Why are we hiding the success / error notification if the dialog is open?

src/features/rtk/actions.js
```diff
@@ -37,3 +41,102 @@ export const toggleAntennaPositionFormat = () => (dispatch, getState) => {
...
+        })
+      );
+    } catch (error) {
+      console.warn('Failed to set saved coordinate:', error);
+
+      // Check if the component is still mounted
+      const currentState = getState();
+      if (!currentState.rtk.dialog.coordinateRestorationDialog.open) {
+        return;
+      }
```
isti115: Why are we hiding the success / error notification if the dialog is open?

src/features/rtk/actions.js
```diff
@@ -37,3 +41,102 @@ export const toggleAntennaPositionFormat = () => (dispatch, getState) => {
...
+
+    const savedCoordinate = {
+      position: antenna.position,
+      positionECEF: antenna.positionECEF,
+      accuracy: survey.accuracy,
+      savedAt: Date.now(),
+    };
+
+    // Check if this exact coordinate is already the latest saved one
+    const savedCoordinates = state.rtk.savedCoordinates[presetId] || [];
```
isti115: Prefer nullish coalescing!
```suggestion
    const savedCoordinates = state.rtk.savedCoordinates[presetId] ?? [];
```

src/features/rtk/selectors.ts
```diff
@@ -159,3 +159,74 @@ export const getSurveyStatus = createSelector(
...
+
+/**
+ * Returns whether there is a saved coordinate for the given RTK preset ID.
+ */
+export const hasSavedCoordinateForPreset = (
+  state: RootState,
+  presetId: string
+): boolean => {
+  const coords = state.rtk.savedCoordinates[presetId];
+  return Boolean(coords && coords.length > 0);
```
isti115:
```suggestion
  return Boolean(coords) && coords.length > 0;
```

src/features/rtk/selectors.ts
```diff
@@ -159,3 +159,74 @@ export const getSurveyStatus = createSelector(
...
+export const getSavedCoordinatesForPreset = (
+  state: RootState,
+  presetId: string
+): RTKSavedCoordinate[] => state.rtk.savedCoordinates[presetId] ?? [];
+
+/**
+ * Returns all saved coordinates as an array of { presetId, coordinates } objects.
+ */
+export const getAllSavedCoordinates = createSelector(
+  (state: RootState) => state.rtk.savedCoordinates,
```
isti115: The dedicated selector mentioned in `RTKStatusUpdater` could be reused here as well.

src/features/rtk/selectors.ts
```diff
@@ -159,3 +159,74 @@ export const getSurveyStatus = createSelector(
...
+ */
+export const getSavedCoordinatesForPreset = (
+  state: RootState,
+  presetId: string
+): RTKSavedCoordinate[] => state.rtk.savedCoordinates[presetId] ?? [];
+
+/**
+ * Returns all saved coordinates as an array of { presetId, coordinates } objects.
+ */
+export const getAllSavedCoordinates = createSelector(
```
isti115: This selector doesn't seem to be referenced anywhere. Is it actually used?

src/features/rtk/selectors.ts
```diff
@@ -159,3 +159,74 @@ export const getSurveyStatus = createSelector(
...
+    Object.entries(savedCoordinates).map(([presetId, coordinates]) => ({
+      presetId,
+      coordinates,
+    }))
+);
+
+/**
+ * Returns the formatted saved coordinate position for a given coordinate.
+ */
+export const getFormattedCoordinatePosition = createSelector(
```
isti115: I'd rather make this a selector that returns a formatter function instead, e.g. `getPreferredSavedRTKPositionFormatter`.

src/features/rtk/RTKCoordinateRestorationDialog.jsx
```diff
@@ -0,0 +1,123 @@
...
+
+const SavedCoordinateItem = connect((state, { coordinate }) => ({
+  formattedPosition: getFormattedCoordinatePosition(state, coordinate),
+}))(({ coordinate, formattedPosition, onClick }) => {
+  const { accuracy, savedAt } = coordinate;
+  const savedDateTime = formatDate(new Date(savedAt), 'yyyy-MM-dd HH:mm:ss');
+
+  return (
+    <ListItem disablePadding>
+      <ListItemButton onClick={() => onClick(coordinate)}>
```
isti115, line 38: In my opinion it's not necessarily immediately obvious that these can be clicked, I would try adding a "load" icon at the end of the list items.

src/features/rtk/selectors.ts
isti115, line 1: I would prefer using the `AppSelector` utility, as I think that it makes type information much more readable for selectors than the automatically inferred monstrosity, but that's probably out of scope for this PR, so I'll go through the file and utilize it where applicable after this has been merged.

src/features/rtk/selectors.ts
```diff
@@ -159,3 +159,74 @@ export const getSurveyStatus = createSelector(
...
+    if (!savedCoordinate) {
+      return undefined;
+    }
+
+    if (isECEF) {
+      const { positionECEF } = savedCoordinate;
+      return positionECEF && Array.isArray(positionECEF)
+        ? `[${(positionECEF[0] / 1e3).toFixed(3)}, ${(
+            positionECEF[1] / 1e3
+          ).toFixed(3)}, ${(positionECEF[2] / 1e3).toFixed(3)}]`
```
isti115: If we insist on this being inline, the following seems like a cleaner alternative to me:
```suggestion
        ? `[${positionECEF.map(c => (c / 1e3).toFixed(3)).join(', ')}]`
```
Otherwise I'd extract it to be reusable with `getFormattedAntennaPosition`.

src/features/rtk/selectors.ts
```diff
@@ -159,3 +159,74 @@ export const getSurveyStatus = createSelector(
...
+/**
+ * Returns the current RTK preset ID.
+ */
+export const getCurrentRTKPresetId = (state: RootState): string | undefined =>
+  state.rtk.currentPresetId;
+
+/**
+ * Returns the coordinate restoration dialog state.
+ */
+export const getCoordinateRestorationDialog = (
```
isti115: I really feel like this should rather be called `getCoordinateRestorationDialogState`.

src/features/rtk/slice.ts
```diff
@@ -8,15 +8,29 @@ import { createSlice, type PayloadAction } from '@reduxjs/toolkit';
 
 import { noPayload } from '~/utils/redux';
 
-import { RTKAntennaPositionFormat, type RTKStatistics } from './types';
+import {
+  RTKAntennaPositionFormat,
+  type RTKStatistics,
+  type RTKSavedCoordinate,
```
isti115, line 14: Sort.
```suggestion
  RTKAntennaPositionFormat,
  type RTKSavedCoordinate,
  type RTKStatistics,
```

src/features/rtk/slice.ts
```diff
@@ -8,15 +8,29 @@ import { createSlice, type PayloadAction } from '@reduxjs/toolkit';
...
+} from './types';
 
 type RTKSliceState = {
   stats: RTKStatistics;
 
+  /** Saved coordinates per RTK preset ID */
+  savedCoordinates: Record<string, RTKSavedCoordinate[]>;
+
+  currentPresetId: string | undefined;
```
isti115: Make it an optional property instead!
```suggestion
  currentPresetId?: string;
```

src/features/rtk/slice.ts
```diff
@@ -8,15 +8,29 @@ import { createSlice, type PayloadAction } from '@reduxjs/toolkit';
...
+
   dialog: {
     open: boolean;
     antennaPositionFormat: RTKAntennaPositionFormat;
     surveySettingsEditorVisible: boolean;
+    /** Dialog for asking user if they want to use saved coordinates */
+    coordinateRestorationDialog: {
+      open: boolean;
+      presetId: string | undefined;
```
isti115: Make it an optional property instead!
```suggestion
      presetId?: string;
```

src/features/rtk/slice.ts
```diff
@@ -95,9 +117,7 @@ const { actions, reducer } = createSlice({
       // gone
       state.stats.satellites = satellites;
 
-      if (state.stats.survey === undefined) {
-        state.stats.survey = {};
-      }
+      state.stats.survey ??= {};
```
isti115, line 121: Nice! 🙂👍

src/features/rtk/slice.ts
```diff
@@ -107,15 +127,95 @@ const { actions, reducer } = createSlice({
...
+    saveCoordinateForPreset(
+      state,
+      action: PayloadAction<{
+        presetId: string;
+        coordinate: RTKSavedCoordinate;
+      }>
+    ) {
+      const { presetId, coordinate } = action.payload;
+
+      state.savedCoordinates[presetId] ||= [];
```
isti115: Why not nullish coalescing assignment? (`??=`)

src/features/rtk/slice.ts
```diff
@@ -107,15 +127,95 @@ const { actions, reducer } = createSlice({
...
+      const existing = state.savedCoordinates[presetId];
+
+      // Check if the latest coordinate is the same as the one we are trying to save
+      if (existing.length > 0) {
+        const latest = existing[0];
+        if (
+          latest &&
+          latest.positionECEF[0] === coordinate.positionECEF[0] &&
+          latest.positionECEF[1] === coordinate.positionECEF[1] &&
+          latest.positionECEF[2] === coordinate.positionECEF[2]
```
isti115: Maybe use `isEqual` from `lodash`? (Performance doesn't really seem to matter here.)

src/features/rtk/slice.ts
```diff
@@ -107,15 +127,95 @@ const { actions, reducer } = createSlice({
...
+        ) {
+          return;
+        }
+      }
+
+      const duplicateIndex = existing.findIndex(
+        (c) =>
+          c.positionECEF[0] === coordinate.positionECEF[0] &&
+          c.positionECEF[1] === coordinate.positionECEF[1] &&
+          c.positionECEF[2] === coordinate.positionECEF[2]
```
isti115: Maybe use `isEqual` from `lodash`? (Performance doesn't really seem to matter here.)

src/features/rtk/actions.js
```diff
@@ -37,3 +41,102 @@ export const toggleAntennaPositionFormat = () => (dispatch, getState) => {
...
+    if (savedCoordinates.length > 0) {
+      const latest = savedCoordinates[0];
+      if (
+        latest &&
+        latest.positionECEF[0] === savedCoordinate.positionECEF[0] &&
+        latest.positionECEF[1] === savedCoordinate.positionECEF[1] &&
+        latest.positionECEF[2] === savedCoordinate.positionECEF[2]
```
isti115: Maybe use `isEqual` from `lodash`? (Performance doesn't really seem to matter here.)

src/features/rtk/slice.ts
```diff
@@ -107,15 +127,95 @@ const { actions, reducer } = createSlice({
...
+      );
+
+      if (duplicateIndex !== -1) {
+        existing.splice(duplicateIndex, 1);
+      }
+
+      existing.unshift(coordinate);
```
isti115, line 154: I personally wouldn't even bother specifically testing for the coordinate being the last saved one, just let it be spliced out and unshifted back in by the general case.

src/features/rtk/slice.ts
```diff
@@ -107,15 +127,95 @@ const { actions, reducer } = createSlice({
...
+      if (existing.length > 5) {
+        existing.pop();
+      }
+    },
+
+    setCurrentRTKPresetId(state, action: PayloadAction<string | undefined>) {
+      state.currentPresetId = action.payload;
+    },
+
+    removeSavedCoordinateForPreset(state, action: PayloadAction<string>) {
```
isti115: Isn't it more correct to call this `removeSavedCoordinatesForPreset`?
If I understand correctly, it removes all the saved coordinates for a given preset.

src/features/rtk/slice.ts
```diff
@@ -107,15 +127,95 @@ const { actions, reducer } = createSlice({
...
   closeSurveySettingsPanel,
+  closeCoordinateRestorationDialog,
+  clearAllSavedCoordinates,
+  removeSavedCoordinateForPreset,
   resetRTKStatistics,
+  saveCoordinateForPreset,
   setAntennaPositionFormat,
+  setCurrentRTKPresetId,
   showRTKSetupDialog,
+  showCoordinateRestorationDialog,
```
isti115, line 196: Sort.
```suggestion
  clearAllSavedCoordinates,
  closeCoordinateRestorationDialog,
  closeRTKSetupDialog,
  closeSurveySettingsPanel,
  removeSavedCoordinateForPreset,
  resetRTKStatistics,
  saveCoordinateForPreset,
  setAntennaPositionFormat,
  setCurrentRTKPresetId,
  showCoordinateRestorationDialog,
  showRTKSetupDialog,
```

src/features/rtk/utils.ts
```diff
@@ -118,3 +120,50 @@ export function formatSurveyAccuracy(
...
+ * @returns True if a valid fix is present, false otherwise.
+ */
+export function hasValidFix(status: Partial<RTKStatistics>): boolean {
+  const hasECEF = Array.isArray(status?.antenna?.positionECEF);
+  const accuracy = status?.survey?.accuracy;
+  const flags = status?.survey?.flags;
+  const surveyedCoordinateValid =
+    typeof flags === 'number' ? (flags & 0b100) !== 0 : false;
+
+  return hasECEF && (surveyedCoordinateValid || typeof accuracy === 'number');
```
isti115: Is the expression in the parentheses deliberately a disjunction? Intuitively a conjunction would make more sense for me, but maybe I'm misunderstanding something.

src/features/rtk/utils.ts
```diff
@@ -118,3 +120,50 @@ export function formatSurveyAccuracy(
...
+  const savedECEF =
+    saved && saved.length > 0 && Array.isArray(saved[0]?.positionECEF)
+      ? saved[0]?.positionECEF.slice(0, 3)
+      : undefined;
+
+  const isSameECEF =
+    incomingECEF && savedECEF
+      ? incomingECEF.length === savedECEF.length &&
+        incomingECEF.every((v, i) => v === savedECEF[i])
+      : false;
```
isti115: Code returning a boolean constant from one side of a ternary can always be simplified into a logical expression. Also, maybe use `isEqual` from `lodash`? (Performance doesn't really seem to matter here.)

src/flockwave/operations.ts
```diff
@@ -226,6 +226,27 @@ export async function startRTKSurvey(
   }
 }
 
+/**
+ * Sets the RTK antenna position on the server by submitting explicit
+ * survey settings that contain a fixed position instead of starting a survey.
+ */
+export async function setRTKAntennaPosition(
+  hub: MessageHub,
+  { position, accuracy }: { position: [number, number, number]; accuracy: number }
```
isti115: Use `Coordinate3D` instead of the inline tuple!

src/flockwave/operations.ts
```diff
@@ -378,6 +399,7 @@ const _operations = {
   setParameter,
   setParameters,
   setRTKCorrectionsSource,
+  setRTKAntennaPosition,
```
isti115, line 476: Sort!
```suggestion
  setRTKAntennaPosition,
  setRTKCorrectionsSource,
```

src/flockwave/queries.js
```diff
@@ -343,7 +343,12 @@ export async function getRTKStatus(hub) {
  * Returns the currently selected RTK data source ID.
  */
 export async function getSelectedRTKPresetId(hub) {
-  const response = await hub.sendMessage({ type: 'X-RTK-SOURCE' });
+  const response = await hub.sendMessage(
+    { type: 'X-RTK-SOURCE' },
+    // Use a larger timeout as the server might be busy reconfiguring the RTK
```
isti115: Sorry, this is really just a nitpick. 🙃
```suggestion
    // Use a longer timeout as the server might be busy reconfiguring the RTK
```

src/i18n/hu.json
```diff
@@ -27,6 +27,15 @@
     "noRTKData": "Nincsenek RTK adatforrások a szerveren",
     "pleaseWait": "Kérem várjon, RTK források betöltése…"
   },
+  "RTKCoordinateRestorationDialog": {
+    "title": "Használod a legutóbbi ismert koordinátát?",
+    "positionLabel": "Pozíció",
+    "accuracyLabel": "Pontosság",
+    "dateLabel": "Mentve",
+    "noSavedCoordinates": "Nincsenek mentett koordináták.",
+    "cancel": "Mégse",
```
isti115: This is unnecessary, see my previous comment on `RTKCoordinateRestorationDialog.jsx`.

src/i18n/hu.json
```diff
@@ -27,6 +27,15 @@
     "noRTKData": "Nincsenek RTK adatforrások a szerveren",
     "pleaseWait": "Kérem várjon, RTK források betöltése…"
   },
+  "RTKCoordinateRestorationDialog": {
+    "title": "Használod a legutóbbi ismert koordinátát?",
```
isti115: This doesn't actually seem to be a translation of the English title.
(I guess that it likely belonged to an earlier iteration of the dialog.)

isti115 (commented):
I tried to provide inline suggestions for all the optional nitpicks, so they can be batch accepted without any extra hassle, as I didn't want to introduce any unnecessary burden requiring further effort on your part. If it still seems too much, I can try to either mark the actually important comments separately or commit some of the trivial changes myself.

isti115 (commented):
I tried to provide inline suggestions for all the optional nitpicks, so they can be accepted without any extra hassle, as I didn't want to introduce any unnecessary burden requiring further effort on your part. If it still seems too much, I can try to either mark the actually important comments separately or commit some of the trivial changes myself.

isti115 (commented):
I tried to provide inline suggestions for all the optional nitpicks, so they can be accepted without any extra hassle, as I didn't want to introduce any unnecessary burden requiring further effort on your part. If it still seems too much, I can try to either mark the actually important comments separately or commit some of the trivial changes myself.

isti115 (commented):
I tried to provide inline suggestions for all the optional nitpicks, so they can be accepted without any extra hassle, as I didn't want to introduce any unnecessary burden requiring further effort on your part. If it still seems too much, I can try to either mark the actually important comments separately or commit some of the trivial changes myself.

src/features/rtk/utils.ts
zarcell: surely accidental

src/features/rtk/RTKCorrectionSourceSelector.jsx
zarcell: Functionally yes

src/features/rtk/RTKSatelliteObservations.jsx
zarcell: If I remember correctly I had issues with re-rendering and this change might have fixed it for me, but I am not sure anymore.

src/store/index.js
zarcell: If I remember correctly the `rtk` was not stored at all so I kept it to only storing what I need. We can change that.

src/features/rtk/slice.ts
zarcell: It was actually only for debugging purposes so I just removed it.

src/features/rtk/RTKCorrectionSourceSelector.jsx
isti115: What is the advantage of the currently present verbose form over the alternative with the nullish coalescing operator?

src/features/rtk/RTKSatelliteObservations.jsx
isti115: @ntamas Can you please comment on this, in case you remember the context from 7388a20?

src/store/index.js
isti115: You're right, `rtk` was in the blacklist previously, so persisting the other keys would be out of scope in this PR, let's discuss it with the rest of the team later!

src/features/rtk/AntennaPositionIndicator.jsx
isti115: For future reference, we have discussed this separately and the reason for removing `focus` as a trigger is to avoid the tooltip popping up when the coordinate restoration dialog is closed, which restores focus to this button.

src/features/rtk/RTKSatelliteObservations.jsx
ntamas: I definitely remember that it was related to chart re-rendering; IIRC the chart was _not_ re-rendering without this when the underlying data was updated. Please be careful with this change.

src/features/rtk/RTKCorrectionSourceSelector.jsx
zarcell: Only the readability for some, but since it really is redundant I will refactor it.

src/features/rtk/utils.ts
```diff
@@ -134,7 +134,8 @@ export function hasValidFix(status: Partial<RTKStatistics>): boolean {
   const surveyedCoordinateValid =
     typeof flags === 'number' && (flags & 0b100) !== 0;
 
-  return hasECEF && (surveyedCoordinateValid || typeof accuracy === 'number');
+  // Consider fix valid only with ECEF position, valid-coordinate flag, and numeric accuracy.
+  return hasECEF && surveyedCoordinateValid && typeof accuracy === 'number';
```
isti115: @ntamas Please review this logic with @zarcell as discussed, by testing using local RTK base stations and remote NTRIP sources as well!

zarcell: Notes for @ntamas when review is done: the LED on the RTK widget shows the surveys current state, which can be misleading when using fixed coordinates from server config or even from this feature. One solution can be a `FIXED` bit in the survey flags that the Live can handle to show exactly what is happening to the user.

ntamas: I'm actually thinkng about redefining the conditions that specify what color the badge of the RTK widget should show such that the "green" LED would appear irrespectively of the survey status if things otherwise seem right.

Basically, thinking from the user's point of view: the user wants to see a green badge there if things "look fine", irrespectively of whether the RTK base station has performed its own survey or is using a fixed coordinate. So, the conditions for the various LED colors could be along the lines of something like:

- if we _know_ that a survey is still in progress, then the LED should be blue (informational)
- if we _can't confirm_ that a survey is still in progress (i.e. no survey info or the survey tells that there is no survey in progress) _and_ either we haven't seen at least one RTK correction message in the last 2 seconds (should be 1 but we add an extra second as a buffer) or we haven't seen an RTK antenna position message in the last 7 seconds (should be 5 but we add an extra 2 seconds as a buffer), then the LED should be red (error)
- if we _can't confirm_ that a survey is still in progress _and_ we have seen at least one RTK correction message in the last 2 seconds _and_ we have also seen RTK antenna position messages in the last 7 seconds, then the LED should be green (things look fine).

This will be the topic of a separate issue or PR but it sidesteps the problem that local RTK base stations usually provide us with survey information while remote NTRIP sources do not provide this information (but they might still be functioning perfectly well).

ntamas (approved):

src/features/rtk/utils.ts
ntamas: Separate issue filed for the remaining points that I raised above: https://github.com/skybrush-io/live/issues/162

ntamas (approved):

PR #107, opened 2025-11-11
UAVs panel: compass calibration operation

volfpeter:
Changes:

- Added a compass calibration operation for selected drones to the UAVs panel's toolbar (`Explore` icon).
- Replaced the `Explore` (compass) icon with the `NearMe` icon in the Environment settings dialog (takeoff heading section).
- Made it possible to pass a timeout option to `messageHub.startAsyncOperation()`.

### DIFF
--- a/src/features/uavs/UAVTestsPanel.jsx
+++ b/src/features/uavs/UAVTestsPanel.jsx
@@ -15,6 +15,7 @@ import Colors from '~/components/colors';
 import { errorToString } from '~/error-handling';
 import { useMessageHub } from '~/hooks';
 
+import { COMPASS_CALIB_TIMEOUT } from './constants';
 import ListItemProgressBar from './ListItemProgressBar';
 
 const tests = [
@@ -22,7 +23,7 @@ const tests = [
     component: 'compass',
     label: 'Calibrate compass',
     type: 'calib',
-    timeout: 90 /* compass calibration may take longer */,
+    timeout: COMPASS_CALIB_TIMEOUT,
   },
   {
     component: 'accel',


src/utils/messaging.js
```diff
@@ -260,8 +266,22 @@ export const turnMotorsOnForUAVs = performMassOperation({
   }),
 });
 
+export const calibrateCompassOnUAVs = performMassOperation(
+  {
+    type: 'OBJ-CMD',
```
isti115: IIRC @ntamas recommended using the more specific [`UAV-CALIB`](https://doc.collmot.com/public/skybrush-protocol-spec/latest/messages/uav.html#_uav_calib_calibrate_a_component_of_one_or_more_uavs) instead of the generic [`OBJ-CMD`](https://doc.collmot.com/public/skybrush-protocol-spec/latest/messages/obj.html#obj-cmd).

src/components/uavs/UAVOperationsButtonGroup.jsx
```diff
@@ -271,6 +273,16 @@ const UAVOperationsButtonGroup = ({
         <>
           {!hideSeparators && <ToolbarDivider orientation='vertical' />}
 
+          <IconButton
+            size={iconSize}
+            disabled={selectedUAVIds.length == 0 || selectedUAVIds.length > 40}
```
isti115: IIRC the number suggested by @ntamas was 20 drones, but this is a completely arbitrary limit anyway, so 40 could be fine too.

volfpeter: Updated in the meantime. It felt very low, but @vasarhelyi clarified it on Discord, so we're back to 20. We now have a tooltip as well, plus the button is now in the same group as details/flash lights.

src/utils/messaging.js
```diff
@@ -84,14 +84,17 @@ const createConfirmationMessage = (operation, uavs, broadcast) => {
...
+  (
+    {
+      type,
+      name,
+      mapper = undefined,
+      reportFailure = true,
+      reportSuccess = true,
+      skipConfirmation = false,
+    },
+    responseHandlerOptions = undefined
```
isti115, line 97: Does assigning `undefined` as default value in a destructuring or function parameter actually do anything? 🤔 Isn't it just the same as not specifying it at all? 🤨 (I mean, we're talking about JS, so of course there might be some really weird edge cases, but I hope that we're not relying on those!)

src/utils/messaging.js
volfpeter: I remembered a `OBJ-CMD`-like command, but to be honest, I checked what message the compass calibration action sends in the UAV details dialog and created an identical message and timeout config, to avoid any potential issues.

We can change to `UAV-CALIB`. Based on the docs, it seems the message would remain the same, we'd only need to change the message type.

src/flockwave/messages.ts
```diff
@@ -1461,7 +1464,7 @@ export default class MessageHub {
    */
   async sendMessage<T = Body>(
     body = {},
-    { timeout = undefined }: { timeout?: number } = {}
+    { timeout = undefined }: TimeoutOptions = {}
```
isti115, line 1467: See https://github.com/skybrush-io/live/pull/107/files#r2514083533

src/utils/messaging.js
volfpeter: It's a second argument, not part of the object destructuring. Even then we don't need the default value, but it doesn't hurt either.

src/flockwave/messages.ts
volfpeter: @isti115 What's the comment?

src/utils/messaging.js
isti115: To me it just feels like noise, the only potential meaning it could carry is that the argument is supposed to be optional, but in that case I'd rather use a temporary JSDoc comment until we eventually migrate this to TypeScript as well.

src/components/uavs/UAVOperationsButtonGroup.jsx
```diff
@@ -188,6 +190,17 @@ const UAVOperationsButtonGroup = ({
             </IconButton>
           </Tooltip>
           {flashLightsButton}
+          <Tooltip content={t('general.commands.calibrateCompass')}>
+            <IconButton
+              size={iconSize}
+              disabled={
+                selectedUAVIds.length == 0 || selectedUAVIds.length > 20
```
ntamas: Let's move the constant (20) to a `constants.ts` file somewhere so it can be raised easily later if there is a need for it.

src/components/uavs/UAVOperationsButtonGroup.jsx
```diff
@@ -188,6 +190,17 @@ const UAVOperationsButtonGroup = ({
...
+              }
+              onClick={() => calibrateCompass()}
```
ntamas, line 201: Here the linter might complain that you shouldn't return stuff from the `onClick` handler so this should be re-written as `() => { return calibrateCompass(); }`. If there are no complaints then it's okay for me.

src/utils/messaging.js
```diff
@@ -260,8 +266,22 @@ export const turnMotorsOnForUAVs = performMassOperation({
...
+  {
+    type: 'UAV-CALIB',
+    name: 'Calibrate compass',
+    mapper: (options) => ({
+      ...options,
+      command: 'calib',
+      args: ['compass'],
+    }),
+  },
+  { timeout: 90 }
```
ntamas: Let's also move this constant to somewhere in a `constants.ts` file.

src/components/uavs/UAVOperationsButtonGroup.jsx
volfpeter: Done. It's in `features/uavs/constants.ts`.

src/utils/messaging.js
volfpeter: Done. It's in `features/uavs/constants.ts`.

src/components/uavs/UAVOperationsButtonGroup.jsx
volfpeter: There's no linter warning here, because the return type of `calibrateCompass()` is `void`.

But thanks for mentioning this. XO keeps breaking for me both in Codium and Zed. At least now I took the time to fix it in Codium (I kind of hate many of its rules though).

volfpeter:
@ntamas I also fixed the compass calibration message. I just noticed while re-testing the PR that it was invalid.

src/features/uavs/constants.ts
```diff
@@ -29,3 +29,14 @@ export const UAV_DETAILS_DIALOG_MIN_WIDTH =
...
+/**
+ * The maximum number of drones compass calibration should be
+ * executed on in a single operation.
+ */
+export const COMPASS_CALIB_UAV_LIMIT = 20;
+
+/**
+ * Timout for the async compass calibration operation.
+ */
+export const COMPASS_CALIB_TIMEOUT = 90;
```
isti115, line 42: Let's use this constant on the [UAV tests panel](https://github.com/skybrush-io/live/blob/feat/uavs-panel-compass-calibration/src/features/uavs/UAVTestsPanel.jsx#L25) as well for consistency!

volfpeter: Good point, thanks! Fixed.

src/components/uavs/UAVOperationsButtonGroup.jsx
ntamas: > I kind of hate many of its rules though

Me too ;-) I wish there was an opinionated linter that happens to align with my preferences in its default configuration. :-D

ntamas:
LGTM now. Feel free to merge.

PR #108, opened 2025-11-12
Fix broken styles after mui v7 upgrade

volfpeter:
_Edit: our `makeStyles()` doesn't understand array class declarations and it also lacks the style registry/ordering features of the mui v6 implementation (meaning we need to use `!important` in cases where we want to override the default style of a component)._


ntamas (approved):
LGTM.

For the record: the problem was that our `makeStyles` replacement did not understand the `[something, '!important']` syntax used in earlier MUI versions.

PR #110, opened 2025-11-18
Feat: light config for show adapt transformation

### LINKED ISSUE #106: Enhance show adapt tool with lighting definition
The skybrush pro backend is about to support  an enhanced `lights` parameter to be used with the takeoff and rth transformations that the show adapt tool calls. We need to update the Live frontend as well to expose the `lights` type and the most important type-specific parameters and send them together with the show adapt xform calls.

volfpeter:
Changes, notes:

- We now use an accordion in the show configurator dialog's sidebar to avoid layout problems (with the dialog footer).
- I added styled `HexColor` components. These are quite similar to how we use `react-colorful` elsewhere in the app, but not exactly the same. Given the lack of time, I did not try to remove code duplication and validate everything. We could create an issue for it.
- @ntamas After merging this PR, Live will require a backend version that supports the new light config models, otherwise show adaptation will fail.

Closes #106

### DIFF
--- a/src/features/servers/selectors.ts
+++ b/src/features/servers/selectors.ts
@@ -146,9 +146,16 @@ export const getServerVersion = (state: RootState): string | undefined =>
 export const getServerVersionValidator = (
   requirement: string
 ): AppSelector<boolean> =>
-  createSelector(getServerVersion, (serverVersion) =>
-    serverVersion ? semver.satisfies(serverVersion, requirement) : false
-  );
+  createSelector(getServerVersion, (serverVersion) => {
+    try {
+      return serverVersion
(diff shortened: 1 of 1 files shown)


volfpeter:
@ntamas I've added the changelog entry and implemented the server version check (+ did minimal cleanup).

The only way to test the UI now is to either change the required server version in `LightConfigurationForm` or to fake a `>=2.40` server version.

The server version checking logic is quite lenient, just a `>=` check. I thought about requiring an exact major version (similar to `^2.40`), but I think that would require maintenance in the future without adding too much value. Let me know what you think.

PR #111, opened 2025-12-03
Feat: collective RTH

volfpeter:
Major changes:

- Collective RTH calculation dialog.
- Collective RTH validation.
- Collective RTH overview panel.
- Pro control group in the Control panel with Suspend and Resume actions.
- Converted most show selectors to TypeScript
- `PromptDialog` is not fully translated

TODOs:

- [x] Clear the stored collective RTH schedule after certain actions (new show loaded, start time changed, ...)
- [x] Fix: use `sendMessage()` instead of `startAsyncOperation()` when sending the show suspend/resume messages
- [x] Show notification on successful show suspend/resume operation
- [x] Ask for confirmation before show suspend/resume (depending on the multi-uav settings)
- [x] (No) Do we need a feature check for the panel? Currently it's set to `collectiveRTH`, but that feature is not registered yet (see  `workbench.js`).
- [ ] Add server version check for this feature?

### DIFF
--- a/src/i18n/en.json
+++ b/src/i18n/en.json
@@ -800,6 +800,10 @@
           "help": "of vertical motion at takeoff & land",
           "label": "Altitude"
         },
+        "altitudeOffset": {
+          "help": "of the net show",
+          "label": "Altitude offset"
+        },
         "horizontalVelocity": {
           "help": "during takeoff & RTH phases",
           "label": "Horizontal velocity"
@@ -812,6 +816,13 @@
           "help": "ensuring exact start of show phase (if nonzero)",
           "label": "Takeoff duration"
         },
+        "takeoffMethod": {
+          "label": "Takeoff method",
+          "type": {
+            "organic": "Organic",
+            "layered": "Layered"
+          }
+        },
         "verticalVelocity": {
           "help": "during takeoff & RTH phases",
           "label": "Vertical velocity"
@@ -819,6 +830,7 @@
       },
       "panel": {
         "lights": "Lights",
+        "safety": "Safety",
         "trajectories": "Trajectories"
       }
     },
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,162 +1,53 @@
-# AGENTS.md
+# Code
 
-This file provides guidelines for agentic coding assistants working on the Skybrush Live codebase.
+## TypeScript
 
-## Build, Lint, and Test Commands
+- Strict typing
+- Use `type` instead of `interface` for type definitions
+- Never cast types using `const x = y as unknown as SomeType` patterns
+- Allow non-null assertions (`!`), but document why it's justified
+- Do not add return type annotation to React components
 
-### Essential Commands
+## Import groups
 
-- `npm run lint` - Run ESLint to check code quality
-- `npm test` - Run all Jest tests
-- `npm test -- path/to/test.test.ts` - Run a single test file
-- `npm test -- -t "test name"` - Run tests matching a pattern
-- `npm start` - Start development web server (http://localhost:8080)
-- `npm run start:electron` - Start Electron desktop app
-- `npm run bundle` - Build for production
-- `npm run lint && npm test` - Run linting and tests together (run this before committing)
+1. External libraries
+2. `@skybrush` libraries
+3. Internal modules with `~/` (alias for `./src/*`)
+4. Relative imports
 
-But you shouldn't try to build the project unless explicitly asked to do so!
+## Exports
 
-### Testing Notes
+- Named exports for utilities and non-React components
+- Default export for React components
 
-- Jest is configured with TypeScript support
-- Test files use `.test.ts` or `.test.tsx` extension
-- Coverage is collected automatically in `coverage/` directory
-- Use `test.skip()` to skip long-running tests (see test/nearestNeighbors.test.ts:82)
+## React
 
-Never run tests unless explicitly asked!
-
-## Code Style Guidelines
-
-### File Format (.editorconfig)
-
-- Encoding: UTF-8
-- Indentation: 2 spaces (no tabs) for JS/JSX/TS/TSX/CSS/LESS
-- Line endings: LF (Unix-style)
-- Final newline: Required at end of files
-
-### TypeScript Rules
-
-- Strict mode enabled
-- Use `type` instead of `interface` for type definitions (ESLint enforced)
-- Use `T[]` for simple array types, `Array<T>` for complex types
-- Use type-only imports: `import type { Foo } from 'bar'` (ESLint enforced)
-- Path alias: `~/*` maps to `./src/*` (e.g., `~/utils/arrays`)
-- Allow `any` type as pragmatic escape hatch for third-party libraries
-- Allow non-null assertions (`!`) when appropriate
-- Unused variables with leading underscore (`_unused`) are allowed
-- Never cast types using patterns like this: `const x = y as unknown as SomeType`.
-- If a named alias exists for a type, use the named alias, that makes the code easier to understand.
-
-### Import Ordering
-
-Import groups:
-
-1. External libraries (e.g., `react`, `lodash-es/has`)
-2. `@skybrush` libraries (e.g. `@skybrush/mui-components`)
-3. Internal modules with `~/` alias (e.g., `~/utils/arrays`)
-4. Relative imports (e.g., `./types`)
-
-Type-only imports are grouped with regular imports, using `type` keyword.
-
-### Naming Conventions
-
-- **Variables/Functions**: camelCase
-- **Components**: PascalCase
-- **Types/Interfaces**: PascalCase (e.g., `type Coordinate2D`)
-- **Constants**: UPPER_SNAKE_CASE (e.g., `NEW_ITEM_ID`, `EMPTY_ARRAY`)
-- **Files**: camelCase for utilities, PascalCase for components
-
-### Export rules
-
-- Use named exports for utilities and non-React components.
-- Use default export for React components.
-
-### Error Handling
-
-- Use custom Error classes when needed (e.g., `class ItemExistsError extends Error`)
-- Include helpful error messages
-- Use try/catch for expected errors; allow errors to bubble up for unexpected ones
-
-### Code Organization
-
-- **Utilities**: Pure functions in `src/utils/` (e.g., arrays, collections, math)
-- **Components**: Organized by feature in `src/components/` and `src/features/`
-- **Hooks**: Custom React hooks in `src/hooks/`
-- **Types**: Shared types in `src/types/` or co-located with usage
-- **Selectors**: Redux selectors in `src/selectors/`
-- **Sagas**: Redux sagas in `src/sagas/`
-
-### React Patterns
-
-- Use function components (declared with `const Comp = () => {}` pattern) with hooks
-- Export default for components, named exports for utilities
-- Use React hooks as needed
+- Use function components declared with `const Comp = () => {}` pattern
 - Prop types defined as TypeScript types
-- If component is connected to the Redux store, create the base non-connected implementation with the desired component name.
-  Then create a connected component with the same name and the "Connected" prefix. Finally default export the connected component.
-- Name components normally. Prefer not to use "Presentation" suffix in component names unless that's the used pattern in code you're editing.
-- Use `useOwnState(props: Props)` hook pattern for deriving state from props within a component.
-- Complex UI sections within a component should be extracted into separate components. They can be in the same file if small, or in separate files if more complex, for better modularity.
+- Redux-connected components: non-connected version with a standard name, connected component with "Connected" prefix
+- Avoid complex UI sections within a component
 
-### Documentation
+## I18n
 
-- Use JSDoc comments for exported functions and complex types
-- Include parameter descriptions with `@param` tags
-- Include return type descriptions with `@returns` tags
-- Add inline comments for non-obvious logic
+- All UI text must be internationalized
+- Never concatenate translated strings, use interpolation in the i18n resource
 
-### Internationalization (i18n)
+## Redux
 
-- **All UI text must be internationalized** - Never hardcode user-facing strings in JSX/TSX files
-- Use translation keys for all text elements, including small UI elements like arrows/separator symbols
-- **Use proper interpolation** - Define translation strings with `{{variable}}` placeholders and pass values as objects: `t('key', { value: ... })`
-- **Don't concatenate translated strings** - Instead of `${t('a')} ${t('b')}`, use a single translation key with multiple parameters
-- **Translation key organization** - Group related translations together in JSON files with consistent naming (e.g., `featureName.section.variableName`)
-- When adding new UI text, always add corresponding translations to `src/i18n/en.json` (or appropriate language files)
+- Redux Toolkit patterns
+- Non-trivial selectors with `createSelector`
 
-### Redux/State Management
+# General
 
-- Use Redux Toolkit patterns
-- Selectors memoized with reselect (`createSelector`)
+**Never** (unless told to):
 
-### Testing Patterns
+- Build or run the project
+- Run or add tests
+- Output explanations or unnecessary code snippets
 
-- Use `describe` to group related tests
-- Use `test` for individual test cases
-- Arrange-Act-Assert pattern preferred
-- Mock external dependencies
-- Use `expect().toBe()`, `expect().toEqual()`, `expect().toThrow()` appropriately
-- Skip long-running tests with `test.skip()`
+**Always**:
 
-### Performance Considerations
-
-- Use `rejectNullish()` for filtering arrays (see src/utils/arrays.ts)
-- Use `EMPTY_ARRAY` and `EMPTY_OBJECT` constants instead of `[]`/`{}`
-- Memoize expensive computations with `memoize-one` or similar
-- Use proper keys in lists (prefer stable IDs over array indices)
-
-### Library Usage
-
-- Use `lodash-es` for utility functions (tree-shakeable ES modules)
-- Use `neverthrow` for Result/Either patterns
-- Use `i18next` for internationalization (translation function: `tt`)
-- Use `react-final-form` for forms with `mui-rff`
-- Use Material-UI (@mui/material) for UI components
-
-## Important Files and Directories
-
-- `src/` - Main source code
-- `test/` - Test files
-- `webpack/` - Webpack configuration
-- `eslint.config.mjs` - ESLint rules (flat config format)
-- `tsconfig.json` - TypeScript configuration
-- `jest.config.ts` - Jest test configuration
-- `babel.config.json` - Babel transpilation config
-
-## General Rules
-
-- Be succint unless explicitly asked to be verbose.
-- If something is unclear, ask for input before you continue.
-- Don't output things or explanations that were not requested.
-- Clearly modularize code, create the correct abstractions. Explore related parts of the codebase and use that as the guideline.
+- Fully understand the task and the related code before getting to work
+- If something is unclear, ask for input
+- Follow clean code principles
+- Be succint
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `PARAM=value` or `PARAM,value` can be used to target all UAVs, and `UAV_ID=PARAM=value`
   or `UAV_ID,PARAM,value` can be used to target a single UAV.
 
+- Added possibility to shift the net show vertically during show adaptation.
+
+- Add takeoff method selection to the show adaptation dialog.
+
+### Changed
+
 - Smaller headings in the Field Notes panel.
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -2780,7 +2780,6 @@
       "integrity": "sha512-q1mjIoW1VX4IvSocvM/vbTiveKC4k9eLrajNEuSsmjymSDEbpGddtpfOoN7YGAqBK3NG+uqo8ia4PDTt8buCYA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": "^18.18.0 || ^20.9.0 || >=21.1.0"
       },
@@ -20058,7 +20057,6 @@
       "integrity": "sha512-atlQQJ2YkO4pfTVQmQ+wvYQwexPDOIgo+RaVcD7gHgzy/IQA+XTyuxNM9M9TVXvttkF7koBHmcwisKdOAf2EcA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@typescript-eslint/eslint-plugin": "8.52.0",
         "@typescript-eslint/parser": "8.52.0",
--- a/src/features/show-configurator/AdaptParametersForm.tsx
+++ b/src/features/show-configurator/AdaptParametersForm.tsx
@@ -3,7 +3,11 @@ import Accordion from '@mui/material/Accordion';
 import AccordionDetails from '@mui/material/AccordionDetails';
 import AccordionSummary from '@mui/material/AccordionSummary';
 import Box from '@mui/material/Box';
+import FormControl from '@mui/material/FormControl';
 import FormGroup from '@mui/material/FormGroup';
+import InputLabel from '@mui/material/InputLabel';
+import MenuItem from '@mui/material/MenuItem';
+import Select from '@mui/material/Select';
 import { useCallback, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 
@@ -15,9 +19,11 @@ import {
   SimpleVelocityField,
 } from '~/components/forms/fields';
 
-import type {
-  OptionalShowAdaptParameters,
-  ShowAdaptParameters,
+import {
+  TAKEOFF_METHODS,
+  type OptionalShowAdaptParameters,
+  type ShowAdaptParameters,
+  type TakeoffMethodType,
 } from './actions';
 import LightConfigurationForm, {
   type LightConfigurationProps,
@@ -26,9 +32,11 @@ import LightConfigurationForm, {
 const defaultAdaptParameters: ShowAdaptParameters = {
   minDistance: 2,
   altitude: 5,
+  altitudeOffset: 0,
   horizontalVelocity: 5,
   verticalVelocity: 1.5,
   takeoffDuration: 0,
+  takeoffMethod: 'layered',
 };
 
 const useStyles = makeStyles((theme) => ({
@@ -51,6 +59,8 @@ function adaptParametersValid(parameters: ShowAdaptParameters): boolean {
   return (
     parameters.minDistance > 0 &&
     parameters.altitude > 0 &&
+    parameters.altitudeOffset >= -100 &&
+    parameters.altitudeOffset <= 100 &&
     parameters.verticalVelocity > 0 &&
     parameters.horizontalVelocity > 0 &&
     parameters.takeoffDuration >= 0
@@ -84,6 +94,9 @@ export function useAdaptParametersFormState(
 ) {
   const [parameters, setParameters] = useState<ShowAdaptParameters>({
     altitude: defaultParameters?.altitude ?? defaultAdaptParameters.altitude,
+    altitudeOffset:
+      defaultParameters?.altitudeOffset ??
+      defaultAdaptParameters.altitudeOffset,
     minDistance:
       defaultParameters?.minDistance ?? defaultAdaptParameters.minDistance,
     horizontalVelocity:
@@ -95,6 +108,8 @@ export function useAdaptParametersFormState(
     takeoffDuration:
       defaultParameters?.takeoffDuration ??
       defaultAdaptParameters.takeoffDuration,
+    takeoffMethod:
+      defaultParameters?.takeoffMethod ?? defaultAdaptParameters.takeoffMethod,
   });
   const [isValid, setIsValid] = useState(adaptParametersValid(parameters));
 
@@ -117,6 +132,19 @@ export function useAdaptParametersFormState(
       const newParameters: ShowAdaptParameters = {
         ...parameters,
         altitude: value,
+      };
+      setParameters(newParameters);
+      setIsValid(adaptParametersValid(newParameters));
+      onChange?.();
+    },
+    [onChange, parameters]
+  );
+  const onAltitudeOffsetChanged = useCallback(
+    (evt: React.ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+      const value = parseDistanceAsMeters(evt.target.value);
+      const newParameters: ShowAdaptParameters = {
+        ...parameters,
+        altitudeOffset: value,
       };
       setParameters(newParameters);
       setIsValid(adaptParametersValid(newParameters));
@@ -163,15 +191,38 @@ export function useAdaptParametersFormState(
     },
     [onChange, parameters]
   );
+  const onTakeoffMethodChanged = useCallback(
+    (
+      event:
+        | React.ChangeEvent<{ value: string }>
+        | { target: { value: string } }
+    ) => {
+      const value: TakeoffMethodType = TAKEOFF_METHODS.includes(
+        event.target.value as TakeoffMethodType
+      )
+        ? (event.target.value as TakeoffMethodType)
+        : 'layered';
+      const newParameters: ShowAdaptParameters = {
+        ...parameters,
+        takeoffMethod: value,
+      };
+      setParameters(newParameters);
+      setIsValid(adaptParametersValid(newParameters));
+      onChange?.();
+    },
+    [onChange, parameters]
+  );
 
   return {
     parameters,
     isValid,
     onMinDistanceChanged,
     onAltitudeChanged,
+    onAltitudeOffsetChanged,
     onHorizontalVelocityChanged,
     onVerticalVelocityChanged,
     onTakeoffDurationChanged,
+    onTakeoffMethodChanged,
   };
 }
 
@@ -188,31 +239,31 @@ const AdaptParametersForm = (props: Props): React.JSX.Element => {
     parameters,
     onMinDistanceChanged,
     onAltitudeChanged,
+    onAltitudeOffsetChanged,
     onHorizontalVelocityChanged,
     onVerticalVelocityChanged,
     onTakeoffDurationChanged,
+    onTakeoffMethodChanged,
   } = props;
   const { t } = useTranslation(undefined, {
     keyPrefix: 'showConfiguratorDialog.adaptParameters',
   });
   const styles = useStyles();
   const [shownSection, setShownSection] = useState<
-    'none' | 'parameters' | 'lights'
-  >('parameters');
+    'none' | 'safety' | 'trajectories' | 'lights'
+  >('safety');
   return (
     <Box>
       <FormGroup>
         <Accordion
-          expanded={shownSection === 'parameters'}
+          expanded={shownSection === 'safety'}
           style={{ margin: 0 }}
           onChange={() => {
-            setShownSection(
-              shownSection === 'parameters' ? 'none' : 'parameters'
-            );
+            setShownSection(shownSection === 'safety' ? 'none' : 'safety');
           }}
         >
           <AccordionSummary expandIcon={<ExpandMore />}>
-            {t('panel.trajectories')}
+            {t('panel.safety')}
           </AccordionSummary>
           <AccordionDetails className={styles.accordionDetails}>
             <SimpleDistanceField
@@ -223,15 +274,6 @@ const AdaptParametersForm = (props: Props): React.JSX.Element => {
               disabled={disabled}
               helperText={t('form.minDistance.help')}
               onChange={onMinDistanceChanged}
-            />
-            <SimpleDistanceField
-              label={t('form.altitude.label')}
-              min={1}
-              max={100}
-              value={parameters.altitude}
-              disabled={disabled}
-              helperText={t('form.altitude.help')}
-              onChange={onAltitudeChanged}
             />
             <SimpleVelocityField
               label={t('form.horizontalVelocity.label')}
@@ -251,6 +293,39 @@ const AdaptParametersForm = (props: Props): React.JSX.Element => {
               helperText={t('form.verticalVelocity.help')}
               onChange={onVerticalVelocityChanged}
             />
+          </AccordionDetails>
+        </Accordion>
+        <Accordion
+          expanded={shownSection === 'trajectories'}
+          style={{ margin: 0 }}
+          onChange={() => {
+            setShownSection(
+              shownSection === 'trajectories' ? 'none' : 'trajectories'
+            );
+          }}
+        >
+          <AccordionSummary expandIcon={<ExpandMore />}>
+            {t('panel.trajectories')}
+          </AccordionSummary>
+          <AccordionDetails className={styles.accordionDetails}>
+            <SimpleDistanceField
+              label={t('form.altitudeOffset.label')}
+              min={-100}
+              max={100}
+              value={parameters.altitudeOffset}
+              disabled={disabled}
+              helperText={t('form.altitudeOffset.help')}
+              onChange={onAltitudeOffsetChanged}
+            />
+            <SimpleDistanceField
+              label={t('form.altitude.label')}
+              min={1}
+              max={100}
+              value={parameters.altitude}
+              disabled={disabled}
+              helperText={t('form.altitude.help')}
+              onChange={onAltitudeChanged}
+            />
             <SimpleDurationField
               label={t('form.takeoffDuration.label')}
               min={0}
@@ -260,6 +335,23 @@ const AdaptParametersForm = (props: Props): React.JSX.Element => {
               helperText={t('form.takeoffDuration.help')}
               onChange={onTakeoffDurationChanged}
             />
+            <FormControl fullWidth variant='filled'>
+              <InputLabel id='takeoff-method-label'>
+                {t('form.takeoffMethod.label')}
+              </InputLabel>
+              <Select
+                disabled={disabled}
+                labelId='takeoff-method-label'
+                value={parameters.takeoffMethod}
+                onChange={onTakeoffMethodChanged}
+              >
+                {TAKEOFF_METHODS.map((value) => (
+                  <MenuItem key={value} value={value}>
+                    {t(`form.takeoffMethod.type.${value}`)}
+                  </MenuItem>
+                ))}
+              </Select>
+            </FormControl>
           </AccordionDetails>
         </Accordion>
         <Accordion
--- a/src/features/show-configurator/actions.ts
+++ b/src/features/show-configurator/actions.ts
@@ -296,13 +296,18 @@ export const adjustHomePositionsToDronePositions =
 type Meters = number;
 type MetersPerSecond = number;
 type Seconds = number;
+export type TakeoffMethodType = 'layered' | 'organic';
+
+export const TAKEOFF_METHODS: TakeoffMethodType[] = ['layered', 'organic'];
 
 export type OptionalShowAdaptParameters = {
-  altitude?: Meters | undefined;
-  minDistance?: Meters | undefined;
-  horizontalVelocity?: MetersPerSecond | undefined;
-  verticalVelocity?: MetersPerSecond | undefined;
-  takeoffDuration?: Seconds | undefined;
+  altitude?: Meters;
+  altitudeOffset?: Meters;
+  minDistance?: Meters;
+  horizontalVelocity?: MetersPerSecond;
+  verticalVelocity?: MetersPerSecond;
+  takeoffDuration?: Seconds;
+  takeoffMethod?: TakeoffMethodType;
 };
 
 export type ShowAdaptParameters = Required<OptionalShowAdaptParameters>;
@@ -407,6 +412,14 @@ export const adaptShow =
         parameters: {
           positions,
           duration: params.takeoffDuration || undefined,
+          method:
+            // TODO: newer servers support both "layered" and "positions",
+            // "positions" being the legacy name of the "layered" method.
+            // We convert to the legacy name here to be on the safe
+            // side. We can remove this in the future.
+            params.takeoffMethod === 'layered'
+              ? 'positions'
+              : params.takeoffMethod,
           lights,
           ...common,
         },
@@ -418,6 +431,16 @@ export const adaptShow =
           ...common,
         },
       },
+      ...(params.altitudeOffset
+        ? [
+            {
+              type: 'shift',
+              parameters: {
+                z: params.altitudeOffset,
+              },
+            },
+          ]
+        : []),
     ];
 
     dispatch(setAdaptResult({ loading: true }));


src/i18n/en.json
```diff
@@ -85,6 +85,51 @@
     "disable": "Disable broadcast",
     "enable": "Enable broadcast for {{time}} seconds"
   },
+  "collectiveRTHDialog": {
+    "action": {
+      "addCollectiveRTH": "Calculate RTH plans"
+    },
+    "description": "Calculate RTH plans for all drones in the show. The RTH plans will be designed to bring the drones home as quickly as possible while avoiding collisions with each other.",
```
vasarhelyi: Suggestion for new wording:

Calculate collective RTH plans for multiple time instants in the show. In case of emergency, collective RTH plans can be triggered to bring the drones home as quickly as possible while avoiding collisions with each other.

src/i18n/en.json
```diff
@@ -85,6 +85,51 @@
...
+    "error": "An error occurred while calculating RTH plans. Please try again.",
+    "existingInvalidRTHPlan": "The swarm has no valid collective RTH plans in the loaded show.",
+    "existingValidRTHPlan": "The swarm has valid RTH plans at {{numPlans}} timestamps in the loaded show.",
+    "loading": "RTH plans are being calculated, please wait. It may take some time.",
+    "parameters": {
+      "form": {
+        "horizontalVelocity": {
+          "help": "during horizontal flight phases",
```
vasarhelyi: "during collective RTH" would be better. Horizontal speed limit is there even for mixed tilted flight paths, not only for horizontal ones.

src/i18n/en.json
```diff
@@ -85,6 +85,51 @@
...
+        "minDistance": {
+          "help": "between drones during RTH",
+          "label": "Minimum distance"
+        },
+        "timeResolution": {
+          "help": "between planned RTH times.",
+          "label": "Time resolution"
+        },
+        "verticalVelocity": {
+          "help": "during vertical flight phases",
```
vasarhelyi: "during collective RTH" would be better. Vertical speed limit is there even for mixed tilted flight paths, not only for vertical ones.

src/i18n/en.json
```diff
@@ -85,6 +85,51 @@
...
   },
+  "collectiveRTHDialog": {
+    "action": {
+      "addCollectiveRTH": "Calculate RTH plans"
+    },
+    "description": "Calculate RTH plans for all drones in the show. The RTH plans will be designed to bring the drones home as quickly as possible while avoiding collisions with each other.",
+    "error": "An error occurred while calculating RTH plans. Please try again.",
+    "existingInvalidRTHPlan": "The swarm has no valid collective RTH plans in the loaded show.",
+    "existingValidRTHPlan": "The swarm has valid RTH plans at {{numPlans}} timestamps in the loaded show.",
+    "loading": "RTH plans are being calculated, please wait. It may take some time.",
```
vasarhelyi: Would be nice to add some kind of a fake progress bar or dynamic icon next to this sentence while RTH plans are being calculated as it might take a longer time and the user should know that he/she is waiting for something.

src/i18n/en.json
```diff
@@ -593,6 +638,10 @@
         "origin": "A show origin must be included or set manually",
         "segments": "The show must be annotated with segment markers",
         "server": "Connection must be established to the server"
+      },
+      "tooltip": {
+        "invalidRTHPlan": "The swarm has no valid collective RTH plans.",
```
vasarhelyi: I would rather write that "The show has no valid ..." not the swarm. Collective RTH plans belong to the show file in general, and specifically at this moment in the show preparation workflow.

src/i18n/en.json
```diff
@@ -593,6 +638,10 @@
...
+        "validRTHPlan": "The swarm has valid RTH plans at {{numPlans}} timestamps."
```
vasarhelyi: I would rather write that "The show has valid ..." not the swarm. Collective RTH plans belong to the show file in general, and specifically at this moment in the show preparation workflow.

Actually I would rephrase to this, I miss the word collective from the sentence: 
`"The show has {{numPlans}} valid collective RTH plans."`

src/i18n/en.json
```diff
@@ -85,6 +85,51 @@
     "disable": "Disable broadcast",
     "enable": "Enable broadcast for {{time}} seconds"
   },
+  "collectiveRTHDialog": {
+    "action": {
+      "addCollectiveRTH": "Calculate RTH plans"
+    },
+    "description": "Calculate RTH plans for all drones in the show. The RTH plans will be designed to bring the drones home as quickly as possible while avoiding collisions with each other.",
+    "error": "An error occurred while calculating RTH plans. Please try again.",
+    "existingInvalidRTHPlan": "The swarm has no valid collective RTH plans in the loaded show.",
```
vasarhelyi: Rephrase to "The loaded show has no valid collective RTH plans."

src/i18n/en.json
```diff
@@ -85,6 +85,51 @@
     "disable": "Disable broadcast",
     "enable": "Enable broadcast for {{time}} seconds"
   },
+  "collectiveRTHDialog": {
+    "action": {
+      "addCollectiveRTH": "Calculate RTH plans"
+    },
+    "description": "Calculate RTH plans for all drones in the show. The RTH plans will be designed to bring the drones home as quickly as possible while avoiding collisions with each other.",
+    "error": "An error occurred while calculating RTH plans. Please try again.",
```
vasarhelyi: add "collective" to RTH

src/i18n/en.json
```diff
@@ -85,6 +85,51 @@
...
+    "existingInvalidRTHPlan": "The swarm has no valid collective RTH plans in the loaded show.",
+    "existingValidRTHPlan": "The swarm has valid RTH plans at {{numPlans}} timestamps in the loaded show.",
```
vasarhelyi: Rephrase to "The loaded show has valid collective RTH plans at {{numPlans}} timestamps".

src/i18n/en.json
volfpeter: Oh, thanks for catching this, it's absolutely necessary.

First I wanted to use a "circular" progress indicator to be consistent with the show configurator dialog. I tried different layouts, I also tried placing the indicator next to the dialog's action buttons, but it always ugly. Then I switched to a progress bar that separated the progress message and the disabled form. It still wasn't nice. Then it came to my mind that we could mimic the upload dialog's UX here and have a status light with a message in the bottom left. This feels pretty decent. The translations changes, you may want to have a look at them: https://github.com/skybrush-io/live/pull/111/changes/620094e617f5a42c6f51cfeee082c2be487c5c75

src/features/collective-rth/actions.ts
```diff
@@ -0,0 +1,109 @@
...
+          : typeof error === 'string'
+            ? error
+            : 'Unknown error type.';
+      dispatch(setResult({ error: errorMessage }));
+    }
+  };
+
+export const saveTransformedShow =
+  (): AppThunk => async (_dispatch, getState) => {
+    const adaptedBase64Show = selectTransformedShowBlob(getState());
```
isti115: These lines (105-108) seem to be identical copies from the `show-configurator` feature, maybe we should use _"transformed"_ here instead of _"adapted"_ in order to stay consistent.

src/features/collective-rth/slice.ts
```diff
@@ -0,0 +1,96 @@
...
+   */
+  showDuration: number;
+
+  /**
+   * Statistics about the collective RTH operation.
+   */
+  stats: StatEntry[];
+
+  /**
+   * The last timestamp at which a collective RTH plan was generated, in seconds.
```
isti115: This doesn't look correct, I guess it's just a copy-paste error.

src/features/show/actions.js
```diff
@@ -472,3 +479,126 @@ export const setOutdoorShowAltitudeReferenceToAverageAMSL =
...
+        semantics: MessageSemantics.ERROR,
+      })
+    );
+  }
+};
+
+export const resumeShow = () => async (dispatch, getState) => {
+  const proceed = await confirmedCollectiveOperation(dispatch, getState, {
+    confirmationMessage: 'show.suspend.confirmation.message',
+    confirmationTitle: 'show.suspend.confirmation.title',
```
isti115: If I assume correctly, these were meant to be updated with `s/suspend/resume` after being duplicated from `suspendShow`.

src/features/collective-rth/CollectiveRTHDialog.tsx
```diff
@@ -0,0 +1,252 @@
...
+  const statusMessage =
+    transformationResult !== undefined
+      ? t('collectiveRTHDialog.status.success')
+      : inProgress
+        ? t('collectiveRTHDialog.status.loading')
+        : error !== undefined
+          ? t('collectiveRTHDialog.status.error')
+          : '';
+
+  const parameteresForm = (
```
isti115: Typo, there's an extra `e`.

src/features/collective-rth/CollectiveRTHDialog.tsx
```diff
@@ -0,0 +1,252 @@
...
+      open={open}
+      title={t('collectiveRTHDialog.title')}
+    >
+      <Box
+        sx={{
+          display: 'flex',
+          flexDirection: 'column',
+          paddingTop: 2,
+          paddingLeft: 4,
+          paddingRight: 4,
```
isti115: Why not use `paddingX` instead?
```suggestion
          paddingX: 4,
```

src/features/collective-rth/CollectiveRTHDialog.tsx
```diff
@@ -0,0 +1,252 @@
...
+      </Box>
+      <DialogActions>
+        <Fade
+          in={
+            inProgress ||
+            transformationResult !== undefined ||
+            error !== undefined
+          }
+        >
+          <Box sx={(theme) => ({ flex: 1, paddingLeft: theme.spacing(1) })}>
```
isti115: Isn't this the same as the following?
```suggestion
          <Box sx={{ flex: 1, pl: 1 }}>
```

src/features/collective-rth/CollectiveRTHDialog.tsx
```diff
@@ -0,0 +1,252 @@
...
+ * The reason for this is to correctly initialize the dialog's state
+ * when it is opened.
+ */
+const CollectiveRTHDialogWrapper = (props: Props) => {
+  const { open, ...rest } = props;
+  return open ? <CollectiveRTHDialog open {...rest} /> : null;
+};
+
+const ConnectedCollectiveRTHDialog = connect(
+  // -- map state to props
```
isti115: There are 211 instances of the comment `// mapStateToProps`, while only two formatted like this.
Let's keep them consistent! (I'd also replace it in the show configurator dialog.)
```suggestion
  // mapStateToProps
```

src/features/collective-rth/CollectiveRTHDialog.tsx
```diff
@@ -0,0 +1,252 @@
...
+  (state: RootState) => ({
+    error: selectTransformationError(state),
+    inProgress: selectTransformationInProgress(state),
+    open: isDialogOpen(state),
+    existingRTHPlanSummary: selectCollectiveRTHPlanSummary(state),
+    transformationResult: selectResult(state),
+  }),
+  // -- map dispatch to props
```
isti115: Similarly to [the comment above](https://github.com/skybrush-io/live/pull/111/changes#r2761668663):
```suggestion
  // mapDispatchToProps
```

src/features/collective-rth/CollectiveRTHDialog.tsx
```diff
@@ -0,0 +1,252 @@
...
+    applyTransformedShow: (show: string): void => {
+      dispatch(loadBase64EncodedShow(show));
+    },
+    closeDialog: (): void => {
+      dispatch(closeDialog());
+    },
+    saveTransformedShow: (): void => {
+      dispatch(saveTransformedShow());
+    },
+  })
```
isti115: Is there any benefit in using the [function form](https://react-redux.js.org/using-react-redux/connect-mapdispatch#defining-mapdispatchtoprops-as-a-function) over the [object shorthand form](https://react-redux.js.org/using-react-redux/connect-mapdispatch#defining-mapdispatchtoprops-as-an-object) here?
I find the following to be much simpler if we don't need the extra flexibility:
```suggestion
  {
    addCollectiveRTH,
    applyTransformedShow,
    closeDialog,
    saveTransformedShow,
  }
```
This also applies to the `ShowConfiguratorDialog`.
⁽¹⁾ https://react-redux.js.org/using-react-redux/connect-mapdispatch#two-forms-of-mapdispatchtoprops

src/features/collective-rth/CollectiveRTHParametersForm.tsx
isti115, line 1: This entire file looks like it could be made much simpler by using `react-final-form` for state management and event handling as well as input parsing and validation.

src/features/show/actions.js
```diff
@@ -472,3 +479,126 @@ export const setOutdoorShowAltitudeReferenceToAverageAMSL =
...
+) => {
+  const baseState = getState();
+  if (selectIsCollectiveRTHTriggered(baseState)) {
+    console.error(
+      'Tried to trigger collective action when collective RTH was already triggered.'
+    );
+    return false;
+  }
+
+  const t = i18n.t;
```
isti115: This pattern of assigning the translation function to a local constant appears several times, but I would discourage it as I don't know if it's explicitly documented to be bound to the `i18n` instance. If that isn't guaranteed, we could potentially run into issues like the following:
```js
const a = new (class {x = 42; f() {return this.x}})
a.f() // 42
const g = a.f
g() // Uncaught TypeError: Cannot read properties of undefined (reading 'x')
```
I'd rather just use it as `i18n.t('...')`.

src/features/show/actions.js
```diff
@@ -472,3 +479,126 @@ export const setOutdoorShowAltitudeReferenceToAverageAMSL =
...
+
+  if (
+    getUAVOperationConfirmationStyle(baseState) !==
+    UAVOperationConfirmationStyle.NEVER
+  ) {
+    const confirmation = await dispatch(
+      showConfirmationDialog(t(confirmationMessage), {
+        title: t(confirmationTitle),
+      })
```
isti115: These calls are problematic, as `babel-plugin-i18next-extract` has no way to figure out what will be passed here. If we really need to accept keys instead of strings, we should use the `tt` function from `~/i18n` to create prepared keys, but I'd rather just move the call to the functions below.

src/features/prompt/PromptDialog.tsx
```diff
@@ -44,44 +45,51 @@ const PromptDialogForm: React.FunctionComponent<PromptDialogFormProps> = ({
...
+        <Form
+          // TODO: Somehow make `fields.SchemaField` use `DialogContent`.
+          formData={initialValues}
+          schema={schema}
+          validator={validator}
+          onSubmit={onSubmit}
+        >
+          <DialogActions style={{ padding: 0 }}>
+            <Button color='primary' type='submit'>
+              {t(submitButtonLabel ?? 'general.action.submit')}
```
isti115: Similar issue to [what I mentioned here](https://github.com/skybrush-io/live/pull/111/changes#r2761709550), but in this case as far as I can see, the `submitButtonLabel` and `cancelButtonLabel` props are never actually utilized, so I would probably just get rid of them for the time being.

src/features/collective-rth/slice.ts
```diff
@@ -0,0 +1,96 @@
...
+   */
+  lastTime: number;
+};
+
+export type TransformationResultOrStatus =
+  | TransformationResult
+  | { error: string }
+  | { loading: true };
+
+export type CollectiveRTHDialogRTHState = {
```
isti115: Do we really need to mention `RTH` twice in this name? 🤔 :slightly_smiling_face:

src/flockwave/schedule.ts
```diff
@@ -0,0 +1,76 @@
+export type TimeSegmentType = 'preparation' | 'rth' | 'slowdown' | 'speedup';
+
+const VALID_TIME_SEGMENT_TYPES = new Set<TimeSegmentType>([
+  'preparation',
+  'rth',
+  'slowdown',
+  'speedup',
+]);
```
isti115, line 10: We usually use enums for similar purpose in the rest of the codebase. For the sake of consistency I'd recommend:
```suggestion
export enum TimeSegmentType {
  PREPARATION = 'preparation',
  RTH = 'rth',
  SLOWDOWN = 'slowdown',
  SPEEDUP = 'speedup',
}

const VALID_TIME_SEGMENT_TYPES = new Set<TimeSegmentType>(
  Object.values(TimeSegmentType)
);
```

src/flockwave/schedule.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  for (const segment of record.schedule) {
+    if (!isTimeSegment(segment)) {
+      return false;
+    }
+  }
+
+  return true;
+}
```
isti115: Lot of already existing type guards follow a different style, so I'd suggest the following:
```suggestion
const isTimeSegment = (data: unknown): data is TimeSegment =>
  // prettier-ignore
  isObject(data)
  && 'type' in data && VALID_TIME_SEGMENT_TYPES.has(data.type as TimeSegmentType)
  && 'startMs' in data && typeof data.startMs === 'number'
  && 'endMs' in data && typeof data.endMs === 'number'
  && (!('params' in data) || data.params === undefined || isObject(data.params));

export const isSchedule = (data: unknown): data is Schedule =>
  // prettier-ignore
  isObject(data)
  && 'schedule' in data
  && Array.isArray(data.schedule)
  && data.schedule.every(isTimeSegment);
```

Or, in case the above is too drastic, at least a minor improvement:
```patch
-  for (const segment of record.schedule) {
-    if (!isTimeSegment(segment)) {
-      return false;
-    }
-  }
+ if (record.schedule.some((segment) => !isTimeSegment(segment))) {
+   return false;
+ }
```

src/features/show/validators.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  let isValid = dronesWithRTHPlan === numDrones && dronesWithoutRTHPlan === 0;
+  if (isValid) {
+    for (const summary of Object.values(plans)) {
+      if (summary?.count !== numDrones) {
+        isValid = false;
+        break;
+      }
+    }
+  }
```
isti115: Isn't this equivalent to the following?
```suggestion
  const isValid =
    dronesWithoutRTHPlan === 0 &&
    dronesWithRTHPlan === numDrones &&
    Object.values(plans).every((summary) => summary?.count === numDrones);
```

src/features/collective-rth/slice.ts
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * The last timestamp at which a collective RTH plan was generated, in seconds.
+   */
+  lastTime: number;
+};
+
+export type TransformationResultOrStatus =
+  | TransformationResult
+  | { error: string }
+  | { loading: true };
```
isti115: I really think that we should make this a proper discriminated union by including an extra dedicated field for the sole purpose of distinguishing between states. For example:
```suggestion
export type TransformationResultOrStatus =
  | { state: 'success', result: TransformationResult }
  | { state: 'failure', error: string }
  | { state: 'loading' }
```

src/features/collective-rth/slice.ts
volfpeter: That name doesn't make sense, I must have renamed this type and didn't notice the second RTH remained in it. I'll fix it.

src/flockwave/schedule.ts
volfpeter: I strongly dislike enums.

src/app.jsx
```diff
@@ -183,6 +184,7 @@ const App = ({ onFirstRender }) => (
       <SafetyDialog />
       <SavedLocationEditorDialog />
       <ShowConfiguratorDialog />
+      <CollectiveRTHDialog />
```
isti115: The list appears to be alphabetically sorted, except for this line.

src/features/collective-rth/actions.ts
volfpeter: Changed it to simply `base64Show`.

src/features/show/actions.js
volfpeter: Thanks! Fixed.

src/features/collective-rth/slice.ts
volfpeter: Yes, thanks, fixed!

src/features/collective-rth/CollectiveRTHDialog.tsx
volfpeter: Fixed.

src/features/show/selectors/trajectory.ts
```diff
@@ -224,9 +224,9 @@ export function makeSelectors(
  *
  */
 export function makeSegmentSelectors(
-  selectSwarm: AppSelector<SwarmSpecification>,
+  selectSwarm: AppSelector<SwarmSpecification | undefined>,
   selectShowSegments: AppSelector<
-    Record<ShowSegmentId, ShowSegment> | undefined
+    Partial<Record<ShowSegmentId, ShowSegment>> | undefined
```
isti115: Let's reuse the [`ShowSegmentsRecord`](https://github.com/skybrush-io/live/blob/f564a67513bcd91a7a164ee89475b60feb0cbc2f/src/features/show/types.ts#L57) type here instead of duplicating the definition!

src/features/collective-rth/CollectiveRTHDialog.tsx
volfpeter: Done.

src/features/collective-rth/CollectiveRTHDialog.tsx
volfpeter: Done.

src/features/collective-rth/CollectiveRTHParametersForm.tsx
volfpeter: I'm not a fan of this lib if I'm honest. I also don't think we should use a lib for everything. I've had enough issues with these (especially `mui-rff`) when doing mui upgrades.

src/features/show/actions.js
volfpeter: I've changed these. You're right, I shouldn't assume that methods are bound.

src/features/show/actions.js
volfpeter: Thanks! I wasn't aware of (or familiar with) this plugin. We now pass the translated string instead.

src/features/prompt/PromptDialog.tsx
volfpeter: Ah, I didn't even intend to do this in the end (after your comments on slack about this). The `prop ??` part should have been outside the `t()` call. Fixed now.

src/features/collective-rth/slice.ts
volfpeter: Fixed now. Thanks for catching it!

src/features/show/validators.ts
volfpeter: Yes, it's much cleaner.

volfpeter: For some reason GH didn't let me commit your suggestion, so I've copied and committed it.

src/features/collective-rth/slice.ts
volfpeter: The types don't overlap, so I think it was okay, but it's definitely cleaner with the discriminator property. I update the type, and the selectors and actions that needed to change.

src/app.jsx
volfpeter: You're right. The feature was strongly related to the show config dialog at the beginning, but that's not relevant here.

src/features/show/selectors/trajectory.ts
volfpeter: It's moved existing code, but you're right. I've updated the typing.

src/store/reducers.ts
```diff
@@ -83,6 +84,7 @@ const dialogsReducer = combineReducers({
   authentication: authenticationReducer,
   deauthentication: deauthenticationReducer,
   dockDetails: dockDetailsDialogReducer,
+  collectiveRTH: collectiveRTHReducer,
```
isti115: The keys appears to be alphabetically sorted, except for this line.

src/utils/formatting.ts
```diff
@@ -325,6 +325,17 @@ export const formatUnixTimestamp = (timestamp: number, naText = '—'): string =
     ? naText
     : formatISO9075(fromUnixTime(timestamp));
 
+/**
+ * Formats a UNIX timestamp in seconds as a time of day text.
+ */
+export const formatTimeOfDay = (
+  timestamp: number | null | undefined,
```
isti115: Is there a reason for allowing `null` and `undefined` here?
If not, let's get rid of them as well as the `isNil` check!

volfpeter: Similar consideration to `formatUnixTimestamp()`. In earlier (wip) versions it sometimes received values that came partly from JS selectors, so I wasn't 100% sure `null` or `undefined` couldn't make its way here. This is no longer the case, it'll receive `NaN`, so making it compatible with `formatUnixTimestamp()` doesn't add anything. It only accepts `number` now.

src/store/reducers.ts
volfpeter: Ah, now I remember: originally the feature was called "emergency RTH" (the branch still has this name), then we did a renaming refactor, that's why we had a couple of sorting issues...

Fixed. Thanks.

src/i18n/en.json
```diff
@@ -85,6 +85,86 @@
     "disable": "Disable broadcast",
     "enable": "Enable broadcast for {{time}} seconds"
   },
+  "collectiveRTHDialog": {
+    "action": {
+      "addCollectiveRTH": "Calculate RTH plans"
+    },
+    "description": "Precalculate collective RTH plans at multiple timestamps in the show. In case of emergency collective RTH brings drones home as quickly as possible without collisions.",
```
isti115:
```suggestion
    "description": "Precalculate collective RTH plans at multiple timestamps in the show. In case of emergency, collective RTH brings drones home as quickly as possible without collisions.",
```

src/i18n/en.json
```diff
@@ -85,6 +85,86 @@
...
+        "horizontalVelocity": {
+          "help": "during collective RTH",
+          "label": "Horizontal velocity"
+        },
+        "minDistance": {
+          "help": "between drones during RTH",
+          "label": "Minimum distance"
+        },
+        "timeResolution": {
+          "help": "between consecutive RTH plans.",
```
isti115:
```suggestion
          "help": "between consecutive RTH plans",
```

src/views/show-control/LargeControlButtonGroup.jsx
```diff
@@ -46,19 +54,49 @@ const useStyles = makeStyles((theme) => ({
...
-      <Divider />
+      <Divider className={classes.divider} />
+      {devModeEnabled && (
+        <>
+          <Accordion
+            expanded={proControlsExpanded}
+            style={{ marginBottom: '2px', boxShadow: 'unset' }}
+            onChange={() => {
+              setProControlsExpanded((value) => !value);
+            }}
```
isti115: Do we actually need to make the state of this accordion controlled?
I think that we should leave it uncontrolled unless we have a reason.

src/views/show-control/LargeControlButtonGroup.jsx
```diff
@@ -46,19 +54,49 @@ const useStyles = makeStyles((theme) => ({
...
+          >
+            <AccordionSummary expandIcon={<ExpandMore />}>
+              <Pro />
+            </AccordionSummary>
+            <AccordionDetails
+              sx={{
+                margin: 0,
+                padding: 0,
+              }}
+            >
```
isti115: Just a nitpick, but in my opinion this is unnecessarily spread out. Also, the rest of the codebase uses the `m` and `p` shorthands in the `sx` prop, so I would consider the following to be more consistent:
```suggestion
            <AccordionDetails sx={{ m: 0, p: 0 }}>
```

src/features/show/validators.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  if (entry.action === 'goToKeepAlt') {
+    if (entry.target.length !== 2) {
+      return false;
+    }
+  } else if (entry.action === 'goToStraight') {
+    if (entry.target.length !== 3) {
+      return false;
+    }
+  }
```
isti115: Why are these conditionals nested?
```suggestion
  if (entry.action === 'goToKeepAlt' && entry.target.length !== 2) {
    return false;
  }

  if (entry.action === 'goToStraight' && entry.target.length !== 3) {
    return false;
  }
```

src/views/show-control/LargeControlButtonGroup.jsx
volfpeter: You're right, we don't need it. I also changed the default state to open, I guess users will want to see the actions by default (at least for now).

src/features/show/actions.js
isti115, line 1: The current implementation only stores the schedule received from `startCollectiveRTH`, but discards the return values of `suspendShow` and `resumeShow`. Is this intentional? To me it seems like this would e.g. allow (the attempt of) triggering another action while one of them is already in progress. Also, this way we can only give feedback for the collective action that's currently being executed if it's RTH, but not suspend and resume.
In case this was a mistake, `collectiveRTHSchedule` in the show slice should probably be renamed to `collectiveActionSchedule`, including it's respective action and selector. If it's intentional, adding an explanatory comment might be a good idea to clarify the reasons behind the decision.

volfpeter: This is intentional. Once collective RTH is triggered, you sit back and wait for the drones to land. There is no way to revert this operation, or suspend and resume it. It may change in the distant future.

Also, the server keeps track of whether suspend, resume, or collective RTH is being executed and responds with an error if you try to trigger something else in the meantime.

We still need to figure out the progress tracking UX for suspend and resume. One for the near future.

src/views/collective-rth/CollectiveRTHPanel.tsx
```diff
@@ -0,0 +1,127 @@
...
+  hasLoadedShowFile: boolean;
+  planSummary: CollectiveRTHPlanSummary;
+};
+
+const useOwnState = ({
+  plans,
+  numDrones,
+  dronesWithoutRTHPlan,
+  dronesWithRTHPlan,
+  isValid,
```
isti115: Maybe sort these?
```suggestion
  dronesWithRTHPlan,
  dronesWithoutRTHPlan,
  isValid,
  numDrones,
  plans,
```

src/views/collective-rth/CollectiveRTHPanel.tsx
```diff
@@ -0,0 +1,127 @@
...
+}: CollectiveRTHPlanSummary) => {
+  const { t } = useTranslation();
+
+  const sortedPlanEntries: CollectiveRTHPlanSummaryItem[] = useMemo(() => {
+    return Object.entries(plans)
+      .map(([time, plan]) => ({
+        time: Number.parseInt(time, 10),
+        maxDuration: plan.maxDuration,
+      }))
+      .sort((a, b) => a.time - b.time);
```
isti115, line 46: I'd prefer `toSorted` to avoid any potential future problems with mutation.
```suggestion
      .toSorted((a, b) => a.time - b.time);
```

src/features/show/validators.ts
```diff
@@ -0,0 +1,175 @@
...
+  let lastTime: number = Number.NEGATIVE_INFINITY;
+  const plans: Record<
+    number,
+    (CollectiveRTHPlanSummaryItem & { count: number }) | undefined
+  > = {};
+
+  for (const drone of drones) {
+    const rthPlan = drone.settings.rthPlan;
+    if (rthPlan === undefined || rthPlan.entries.length === 0) {
+      dronesWithoutRTHPlan++;
```
isti115, line 112: Do we need to count these separately? I feel like this just introduces an opportunity for inconsistency if the value could be just as easily computed by doing `numDrones - dronesWithRTHPlan` on the `CollectiveRTHPanel`.

src/views/collective-rth/CollectiveRTHPanel.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+    if (dronesWithRTHPlan === numDrones && !isValid) {
+      return {
+        message: t('collectiveRTHPanel.error.invalidPlan'),
+        severity: 'error',
+      };
+    }
+
+    return undefined;
+  }, [dronesWithRTHPlan, dronesWithoutRTHPlan, numDrones, isValid, t]);
```
isti115: Sort?
```suggestion
  }, [dronesWithRTHPlan, dronesWithoutRTHPlan, isValid, numDrones, t]);
```

src/views/collective-rth/RTHPlanDetails.tsx
isti115, line 1: Why doesn't this use the dedicated `Table` component from Material UI? 🤔
(If there's need for virtualization, it's perfectly [compatible with `react-virtuoso`](https://mui.com/material-ui/react-table/#virtualized-table).)

---

Also, how many plans are expected for a show?
Am I very far off to asume e.g. one every 5 seconds in a 10 minute long show ➡️ ≈120?
That doesn't seem big enough for me to warrant virtualization, especially since the rows will only be re-rendered if either the plans or the start time changes, thus not that often.

src/views/collective-rth/CollectiveRTHPanel.tsx
volfpeter: The array is created right here...

src/features/show/validators.ts
volfpeter: Given the original representation, I think it's best to count both.

src/views/show-control/ShowConfiguratorButton.tsx
```diff
@@ -71,16 +72,26 @@ const PREREQUISITES: ReadonlyArray<
...
   base64Blob?: string;
+  devModeEnabled: boolean;
   show: ShowData | undefined;
   partialShow: Partial<ShowData>;
   // TODO: This should probably be a `ThunkActionDispatch`, but that doesn't
   //       seem to be reexported from `redux-thunk` via `@reduxjs/toolkit`...
   showDialogAndClearUndoHistory: (data?: ShowData) => void;
+  showCollectiveRTHDialog: () => void;
   status: Status;
+  rthPlanSummary: CollectiveRTHPlanSummary;
```
isti115: Sort?
```suggestion
  base64Blob?: string;
  devModeEnabled: boolean;
  partialShow: Partial<ShowData>;
  rthPlanSummary: CollectiveRTHPlanSummary;
  show: ShowData | undefined;
  showCollectiveRTHDialog: () => void;
  // TODO: This should probably be a `ThunkActionDispatch`, but that doesn't
  //       seem to be reexported from `redux-thunk` via `@reduxjs/toolkit`...
  showDialogAndClearUndoHistory: (data?: ShowData) => void;
  status: Status;
```

src/views/show-control/ShowConfiguratorButton.tsx
```diff
@@ -71,16 +72,26 @@ const PREREQUISITES: ReadonlyArray<
...
-const ShowConfiguratorButton = (props: Props): React.JSX.Element => {
-  const { show, showDialogAndClearUndoHistory, status } = props;
+const ShowConfiguratorButton = (props: Props) => {
+  const {
+    devModeEnabled,
+    show,
+    showDialogAndClearUndoHistory,
+    showCollectiveRTHDialog,
+    status,
+    rthPlanSummary,
```
isti115: Sort?
```suggestion
    devModeEnabled,
    rthPlanSummary,
    show,
    showCollectiveRTHDialog,
    showDialogAndClearUndoHistory,
    status,
```

src/views/collective-rth/RTHPlanDetails.tsx
volfpeter: I used the mui table at some point, although without virtualization if I remember correctly (not sure if I ever committed that). The aim here was to make this table similar to the "UAV list".

Early on we assumed virtualization was unnecessary, but it turned out RTH plan calculation is not as slow as expected, so Tamás decided that we should do virtualization (assuming an RTH plan will be calculated for every second of the show segment).

src/views/show-control/ShowConfiguratorButton.tsx
```diff
@@ -149,6 +160,36 @@ const ShowConfiguratorButton = (props: Props): React.JSX.Element => {
...
+              size='large'
+              color={
+                rthPlanSummary.numDrones === 0
+                  ? 'default'
+                  : rthPlanSummary.isValid
+                    ? 'success'
+                    : 'warning'
+              }
+              onClick={(evt) => {
+                evt.stopPropagation();
```
isti115, line 185: Isn't there a dedicated component / property for setting secondary actions for list items that avoids the overlapping event handlers?

src/views/show-control/ShowConfiguratorButton.tsx
isti115, line 1: With the current hierarchy the collective RTH dialog inherits the `PREREQUISITES` set for the show adapt dialog, which is almost correct, except for the requirement of having set a show origin, which is not necessary for generating cRTH plans if I understand correctly. Their `disabled` states should probably be separate, which might even be the case by default when using a dedicated secodary list item action.

src/views/show-control/ShowConfiguratorButton.tsx
volfpeter: There was, it's deprecated and the replacement prop (`secondaryAction`) doesn't exist on `ListItemButton` if I'm not mistaken.

src/views/show-control/ShowConfiguratorButton.tsx
volfpeter: This is technically correct, but it's an implementation detail and in practice I don't think it matters if I'm honest. I actually think it's best to use the same logic.

src/views/collective-rth/CollectiveRTHPanel.tsx
isti115: Yes, but it might as well be refactored later, such that it will not be... (That's why I mentioned _potential **future** problems_. :wink:)
(The only downside of `toSorted` is the performance hit, but it will not be a repeated operation in this case.)

volfpeter: I know you unconditionally hate mutation. I agree it's best avoid it in most cases, but this is not one of them. The array is created right there, there are **no** potential future problems. If you replace the prop with something totally different, then this entire `useMemo()` will need to be rewritten...

Feel free to add more thoughts, but from my end, I really don't think it's worth commenting about anymore, so I'll resolve the comment.

src/features/show/selectors/index.ts
```diff
@@ -0,0 +1,7 @@
+export * from './core';
+export * from './js-core';
+export {
+  selectCollectiveRTHPlanSummary,
```
isti115, line 4: Is there a reson why this one is specifically re-exported, while `selectCollectiveRTHSchedule` and `selectIsCollectiveRTHTriggered` are referenced from `selectors/rth` directly?

src/features/collective-rth/CollectiveRTHDialog.tsx
```diff
@@ -0,0 +1,243 @@
...
+          alignItems: 'center',
+        }}
+      >
+        <Button
+          color='primary'
+          onClick={() => addCollectiveRTH(parametersFormState.parameters)}
+        >
+          {t('collectiveRTHDialog.action.addCollectiveRTH')}
+        </Button>
+      </Box>
```
isti115: Why hide the button completely while waiting for a response instead of just putting it in a [loading state](https://mui.com/material-ui/react-button/#loading-2)?

Also, the styling / layout can be simpler:
```suggestion
      <Button
        color='primary'
        loading={inProgress}
        loadingPosition='start'
        onClick={() => addCollectiveRTH(parametersFormState.parameters)}
        sx={{ margin: 'auto' }}
      >
        {t('collectiveRTHDialog.action.addCollectiveRTH')}
      </Button>
```

src/features/collective-rth/CollectiveRTHDialog.tsx
```diff
@@ -0,0 +1,243 @@
...
+    </Box>
+  );
+
+  return (
+    <DraggableDialog
+      fullWidth
+      disableEscapeKeyDown={inProgress || transformationResult !== undefined}
+      maxWidth='sm'
+      open={open}
```
isti115, line 108: Improve consistency with other dialogs by closing when the backdrop is clicked.
```suggestion
      maxWidth='sm'
      onClose={closeDialog}
      open={open}
```

src/features/collective-rth/CollectiveRTHDialog.tsx
```diff
@@ -0,0 +1,243 @@
...
+/**
+ * Wrapper that only renders the dialog when it is open.
+ *
+ * The reason for this is to correctly initialize the dialog's state
+ * when it is opened.
+ */
+const CollectiveRTHDialogWrapper = (props: Props) => {
+  const { open, ...rest } = props;
+  return open ? <CollectiveRTHDialog open {...rest} /> : null;
+};
```
isti115: Let's reduce clutter by simplifying the arrow function to use an expression body:
```suggestion
const CollectiveRTHDialogWrapper = ({ open, ...rest }: Props) =>
  open ? <CollectiveRTHDialog open {...rest} /> : null;
```

src/features/collective-rth/actions.ts
```diff
@@ -0,0 +1,118 @@
...
+import { setResult } from './slice';
+
+export type OptionalCollectiveRTHParameters = {
+  minDistance?: number;
+  timeResolution?: number;
+  horizontalVelocity?: number;
+  verticalVelocity?: number;
+};
+
+export type CollectiveRTHParameters = Required<OptionalCollectiveRTHParameters>;
```
isti115, line 18: Defining these in the opposite direction would seem much more natural to me:
```suggestion
export type CollectiveRTHParameters = {
  minDistance: number;
  timeResolution: number;
  horizontalVelocity: number;
  verticalVelocity: number;
};

export type OptionalCollectiveRTHParameters = Partial<CollectiveRTHParameters>;
```

isti115:
One more observation I had was the usage of the _"select..."_ prefix when defining selectors.
The only precedent for this that I am able to find is in the show adaptation feature, otherwise the codebase uses _"is..."_ / _"has..."_ / _"does..."_ and similar prefixes for boolean selectors and _"get..."_ for other return types.
I think that it would be worth renaming the ones currently added as well as those included previously as part of #57 for the sake of consistency.

src/features/collective-rth/actions.ts
volfpeter: Not to me.

src/features/collective-rth/CollectiveRTHDialog.tsx
volfpeter: I find object destructuring in function signatures really ugly and harder to read. I strongly disagree on your "clutter" comment. In some cases I still use it, because I see it's a common pattern in Live, but I still don't like it. But in order to avoid an argument, I'll commit your suggestion.

src/features/collective-rth/CollectiveRTHDialog.tsx
volfpeter: This is not consistent with the show config dialog, which is the closest to this one in functionality: a dialog triggers a long calculation on the backend, and losing result by accidentally closing the dialog can be an issue.

volfpeter: Hm, I double-checked this to be sure, and it seems the show adapt dialog can now also be closed like this, so I'll commit this suggestion. I wasn't aware of that change.

src/views/show-control/ShowConfiguratorButton.tsx
ntamas: Let's keep things simple. In practice, it's very unlikely that the user does not have a show origin at all; typically this happens the first time when they install Live, and after every configuration reset with the dedicated hotkey, but most users do not even know that such a hotkey exists. So, in practice, users have a show origin 99.9999% of the time and it makes no sense to complicate things just to get the remaining 0.0001% corner case correct. (Mathematicians would probably disagree but they are not being paid by the number of features delivered ;-)).

src/views/collective-rth/RTHPlanDetails.tsx
ntamas: It's fine for me either way; given that we already have a working implementation I don't really see much business value in rewriting it.

As for the original question about the number of plans: originally I had this very ambitious goal of generating a plan for every second of the show. Nowadays I'm starting to think that maybe every 5 seconds is enough. The problem is not really the time that we need to calculate the RTH plan but the extra amount of data that we need to transfer to the drones, especially considering that the current firmware _reads the entire show file in memory_ to avoid having to do any SD card operations on the main thread of the firmware during the show. If we have a plan for every second, this might increase the memory usage significantly while not really bringing more benefits than having a plan for every 5th second. The time that the user needs to wait before the actual operation starts is dominated by the preparation period during which we need to broadcast the information that a collective RTH operation is about to take place _at some nearby point in the future_. We need to keep on broadcasting this information for at least 5 seconds to ensure that all the drones get it. If we only have an RTH plan every 5 seconds, it means an average waiting time of 5+2.5 = 7.5 seconds compared to 5+0.5 = 5.5 seconds if we have an RTH plan for every second -- not really a significant gain.

isti115: The value I see in simplifying this comes from the ease of maintainance later. If we can drop virtualization and switch to a native MUI table, I'm certain that the implementation would become **much** cleaner, to the point of it being trivial, and since this feature is just about to undergo a lot of testing, there's no better point in time to get rid of the unnecessary complexity than right now. If we keep things this way, it will be oddly sticking out of the codebase and such unfamiliar / strange parts often end up being avoided in terms of modifications later on in order to not break it, as we don't remember why they're like that in the first place. Since this component is so basic, throwing it away and completely reimplementing it later is also possible, but I really don't like seeing the wheel (or in this case, the table) being reinvented with no clear reason.

_Edit: It looks like I will have to walk this back a little bit, as I was convinced that the reduction would be at least 50% in terms of LoC, but now that [I have actually tried](https://gist.github.com/isti115/178ff65ed1caee96b26500ae6f31a864) for the sake of proving my case, I see that it only gets down to around 55% of the original. I also realize that this is such an independent and insignificant little component that even though I still believe in the reasons I outlined above, they might not be worth applying here..._

src/views/collective-rth/RTHPlanDetails.tsx
```diff
@@ -0,0 +1,219 @@
...
+    gridTemplateRows: 'max-content 1fr',
+    gap: theme.spacing(1),
+    height: '100%',
+  },
+}));
+
+type Props = {
+  plans: CollectiveRTHPlanSummaryItem[];
+  firstTime: number | undefined;
+  lastTime: number | undefined;
```
isti115: These props seem to be unused and unnecessarily passed from `CollectiveRTHPanel`.

volfpeter: Leftover from earlier versions, removed.

src/components/Countdown.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { useTranslation } from 'react-i18next';
+import { formatDuration } from '~/utils/formatting';
+
+const INTERVAL_MS = 250;
+
+type CountdownProps = {
+  seconds: number;
+};
+
+const Countdown = ({ seconds }: CountdownProps) => {
```
isti115, line 11: This component _seems_ like it might overlap with the functionality provided by the `react-timeago` package that's already used in several places, it might be worth seeing if it could be used instead.

src/views/show-control/ShowConfiguratorButton.tsx
volfpeter: For the record, I'm not sure the behavior you mentioned ever worked: `UAVLogsPanel` relied on it, and when the user clicked the download/save button on a log item, the click handler was triggered twice (in v2.11). It would be implicit and fragile anyway.

PR #115, opened 2025-12-16
Feat: color support for sparks light config

volfpeter:


vasarhelyi:
I tested it, works OK, thanks!

PR #117, opened 2026-01-09
"original" light config support

volfpeter:
~~Pending a backend fix, ping @vasarhelyi~~ Pending backend release.

Maybe we should discuss the order in the light config dropdown. Now the first item is "Off", then the rest are in alphabetical order. But maybe a full alphabetical order would be better, especially because the default is "Default", which also happens to be the first one in alphabetical order right now :slightly_smiling_face: 

See `lightEffectTypes` in `LightConfigurationForm.tsx`. Let me know what you think.


vasarhelyi:
ABC order is OK of course.

PR #124, opened 2026-02-16
Fix: uav ID changes in UAV properties dialog (tests and logs panels)

volfpeter:
## `UAVTestsPanel`

- Fix hook dependency lists to trigger tests on the correct UAV
- Fix task progress reporting issues when the UAV ID changes back and forth
- Fix status light when the UAV ID changes

Note: async task/test progress tracking is still not perfect, just better than it used to be. We could track progress for multiple UAVs simultaneously, and immediately show the progress and status when the user switches drones and triggers tests for several drones one by one. This would complicate the component quite a bit, and we may decide to implement a more general solution, so I'd rather delay fixing this use-case.

## `UAVLogsPanel`

- The download and save actions were triggered twice when the user clicked the icon and not the list item itself. Reason: the click handler was registered on both of these components, and there were no guards to avoid repeated event handling.

### DIFF
--- a/src/features/uavs/UAVLogsPanel.jsx
+++ b/src/features/uavs/UAVLogsPanel.jsx
@@ -10,7 +10,7 @@ import Typography from '@mui/material/Typography';
 import isNil from 'lodash-es/isNil';
 import prettyBytes from 'pretty-bytes';
 import PropTypes from 'prop-types';
-import React, { memo, useCallback } from 'react';
+import { memo, useCallback } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { useAsyncRetry } from 'react-use';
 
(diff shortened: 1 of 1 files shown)


src/features/uavs/UAVLogsPanel.jsx
```diff
@@ -159,12 +159,7 @@ const UAVLogListItem = ({ id, kind, size, timestamp, uavId }) => {
         secondary={secondaryComponent}
       />
       <ListItemSecondaryAction>
-        <IconButton
```
ntamas, line 162: Where did the `onClick` handler go?

volfpeter: The icon and the list item had the same click handler. If you clicked the button, the event handler was executed once, as you'd expect. But if you clicked the icon, it was executed twice (once through the icon, and once through the list item). Removing the click handler from the icon button was the trivial fix.

PR #131, opened 2026-02-25
Feat/show adapt alt shift + takeoff method

vasarhelyi:
This PR reorganizes the show adapt parameters into 3 sections, Safety, Trajectories and Lights and adds two new params:

- [x] "Altitude offset" with which users can shift the net show vertically
- [x] "Takeoff method" to select from layered or organic takeoffs.

Needs pro backend with `skybrush-ext-studio` version > 1.4.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add takeoff method selection to the show adaptation dialog.
 
+### Changed
+
+- Smaller headings in the Field Notes panel.
+
 ### Fixed
 
 - Fixed UAV changing issues on the Tests panel of the UAV properties dialog.
--- a/src/views/field-notes/FieldNotesPanel.jsx
+++ b/src/views/field-notes/FieldNotesPanel.jsx
@@ -59,6 +59,30 @@ const useStyles = makeStyles((theme) => ({
         opacity: 1,
       },
     },
+
+    '& .cm-s-easymde .cm-header-1': {
+      fontSize: 'calc(1.2rem + 0.4vw)',
+    },
+
+    '& .cm-s-easymde .cm-header-2': {
+      fontSize: 'calc(1.2rem + 0.3vw)',
+    },
+
+    '& .cm-s-easymde .cm-header-3': {
+      fontSize: 'calc(1.2rem + 0.2vw)',
+    },
+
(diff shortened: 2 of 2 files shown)


src/features/show-configurator/actions.ts
```diff
@@ -407,6 +412,14 @@ export const adaptShow =
         parameters: {
           positions,
           duration: params.takeoffDuration || undefined,
+          method:
+            // TODO: newer servers support both "layered" and "positions",
```
vasarhelyi, line 416: This TODO can be eliminated with using latest commit in the `skybrush` repo as of today, resolving to `method: params.takeoffMethod,` simply.

volfpeter: I'd wait a couple of server releases before removing this. We can create an issue to remove this.

ntamas: We cannot make assumptions about which version of the server the user is going to use. Some clients are likely to be conservative and they tend to stay on older versions for a while. Keep this for at least the current major version of Live.

src/i18n/en.json
```diff
@@ -622,6 +622,10 @@
           "help": "of vertical motion at takeoff & land",
           "label": "Altitude"
         },
+        "altitudeOffset": {
+          "help": "of the net show",
```
ntamas, line 626: I'm not 100% satisfied with this as this might not play nicely with other languages Let's not change this for now but in the future it's probably better to use full sentences in help texts. (I see that there are quite a few other examples of this by now).

AGENTS.md
ntamas, line 1: Changes in `AGENTS.md` are unrelated to the main topic of this PR. I'd rather have this as a separate PR.

src/i18n/en.json
vasarhelyi: You are right, I just hate repeating the beginning of the sentence... All of these can be modified if needed, but then all texts will be longer and redundant. But I am happy to update all if you say so.

ntamas: No need to do it now, just mentioned that this is bad UX and it should be avoided in the future. (Your personal inconvenience for repeating the beginning of the sentence is outweighed by the benefits of making the text easier to interpret by the end users).

AGENTS.md
volfpeter: Ok. I usually treat this similarly to readme, changelog, etc., so I make changes/cleanup in PRs, it's not really a code or feature change, and I wanted to avoid creating a separate PR just for this. I've reverted the change.

PR #132, opened 2026-02-26
Refactor: migrate to react-hot-toast

### LINKED ISSUE #99: Replace react-toast-notifications with react-hot-toast
`react-toast-notifications` is not maintained (it's been abandoned years ago), and it only supports React 17.

volfpeter:
Closes #99 

Changes:

- Replace `react-toast-notifications` with `react-hot-toast`
- Don't require redux to show notifications
- Use the standard MUI `Alert` for notifications (semi-transparent + fixed 420px width, adjusted to the suspend/resume notifications)
- Some cleanup and typing fixes (in `Notification`)
- Removed notification `header` support (it wasn't used)

Questions:

- Simplify package, don't use `redux` terminology, maybe also rename to `notification` to make it easier to find?

### DIFF
--- a/src/features/snackbar/Toast.tsx
+++ b/src/features/snackbar/Toast.tsx
@@ -23,7 +23,6 @@ const StyledAlert = styled(Alert)({
   '> .MuiAlert-message': {
     flex: 1,
   },
-  opacity: 0.85,
 });
 
 const semanticsToSeverity: Record<MessageSemantics, AlertColor> = {


src/views/features/FeaturePanelToolbar.jsx
```diff
@@ -83,14 +83,12 @@ export default connect(
           }
         });
 
-        dispatch(
-          showSuccess(
-            // TODO: Use the `pluralize` package?
-            `Successfully imported ${counter} feature${counter > 1 ? 's' : ''}`
-          )
+        showSuccess(
+          // TODO: Use the `pluralize` package?
```
ntamas, line 87: Probably we would localize this eventually "the right way", and then the plural would be handled by `react-i18next`

PR #136, opened 2026-03-03
Enable show origin transformations through interactions with the takeoff markers

### LINKED ISSUE #79: Select, grab and move takeoff positions together with gross show hull on main map
On the main map one can fine-tune the position of the entire takeoff grid with selecting the gross convex hull and left-dragging it, which moves the takeoff positions as well after releasing the left mouse button. There are some UX enhancements that would be great for this action:

- [ ] it would be nice to move the takeoff positions as well in an animated way while left-dragging the gross convex hull, as one usually adjusts the takeoff positions to the drones, and not the convex hull. It might run into a speed issue but worth a try with e.g. 5000 drones how it goes
- [ ] It would be nice if the takeoff positions themselves would be selected as well visually, when one selects the gross convex hull
- [ ] this "gross convex hull selects takeoff positions" coupled selection could be used on the new show-adapt map as well, there extending the selection to the net show hull as well
- [ ] it is essential to be able to grab and left-drag the takeoff positions themselves together with the convex hull, because when adjusting the takeoff positions to the actual drone positions the convex hull border might be far away in a fine zoom, it would be much more convenient to zoom onto the takeoff positions and left-drag those to the actual drone positions

isti115:
This is a very minimal working prototype that needs more testing / consideration to decide what other special cases need to be handled.

Can be experimented with at: http://live.skybrush.donko.hu/(feat/takeoff-position-selection)

For example currently there's something weird going on if the user tries to manipulate the takeoff positions after a show is unloaded.

Also, the behavior should probably be different in the generic mission context. Maybe for the time being we should just completely disable the interactions unless we are in show mode and just deal with the rest later.

### DIFF
--- a/src/views/map/MapView.jsx
+++ b/src/views/map/MapView.jsx
@@ -38,7 +38,7 @@ import { getSelectedTool } from '~/features/map/tools';
 import { updateMapViewSettings } from '~/features/map/view';
 import { addNewMissionItem } from '~/features/mission/actions';
 import { getGeofencePolygonId } from '~/features/mission/selectors';
-import { getSelection } from '~/features/selection/selectors';
+import { getVirtualSelection } from '~/features/selection/selectors';
 import {
   addToSelection,
   removeFromSelection,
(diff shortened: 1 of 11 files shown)


src/views/map/MapView.jsx
```diff
@@ -732,6 +744,7 @@ const MapView = connect(
     selectedFeatures: getSelectedFeatureIds(state),
     selectedTool: getSelectedTool(state),
     selection: getSelection(state),
+    takeoffPositionIds: getMissionMapping(state).map((_, i) => String(i)),
```
ntamas: Won't this cause an unnecessary re-render every time the state changes, even if the mission mapping itself does not change? If I understand correctly, this produces a new array for every invocation.

isti115: This was just a quick and dirty version to check if we should even try to stick to this approach at all, but yes, it probably would result in a perfornance hit. I'll resolve this anyway, as due to @vasarhelyi's discovery about loading shows with different number of drones I think that we should switch to a completely different implementation.

isti115:
@vasarhelyi Can you give this another look and see if you can spot any remaining issues?
https://live.skybrush.donko.hu/(feat/takeoff-position-selection)

vasarhelyi:
Tested now with latest commit, I could not find any more bugs, nice job, thanks, works very smoothly and definitely a much better UX than before.

PR #149, opened 2026-03-18
Selection groups

volfpeter:
Prototype for group selection handling. UI/UX is still very much up for debate (changes can potentially invalidate the underlying implementation).

Changes:

- Added a selection groups button next to the UAV status summary in the app's header. Users can create, update, delete, and select selection groups from here. This is likely a temporary location and maybe also a temporary UI.
- Selection groups are stored in the `selection` slice in a sorted collection (by ID). IDs are numeric and start from 1. When a new group is saved, it is assigned the first unused numeric ID.
- Selection groups have editable names. The default name is "Group {id}". There's no selection group editor UI yet (it would be a single text input somewhere), so there is no way to actually edit the group name (although the redux state supports it). A dialog felt overkill. Let's discuss this.
- I added a hotkey for selecting selection groups: `g{id}`. There's no hotkey yet for example for creating or updating a group (from the current selection). I'll add these features when it's confirmed we're on the right path.


src/store/index.js
```diff
@@ -62,6 +62,7 @@ const persistConfig = {
     'log',
     'logDownload',
     'messages',
+    'selection',
```
ntamas: If we blacklist the entire `selection` slice, the groups are lost when the user closes the app and opens it again. While I know that groups are meant to be temporary, this probably goes against the user's expectations so I would probably keep the `selection` slice in the persisted state.

ntamas (approved):
Just one small comment. I'm mostly happy with the UX; this is probably not the final version but all the small tweaks that I plan to try or add can be implemented after this PR is merged (or even in upcoming Live versions). So let's deal with the single comment that I added and then this can be merged.

src/store/index.js
volfpeter: I felt this is the safe choice, because reloaded groups may not be valid. If we identify the right interactions that possibly make selection groups invalid, then I'll also be in favor of persisting this state.

ntamas: I think that the app is pretty robust in handling cases when you have something in the selection that does not exist any more. The _current_ selection used to be persisted for a long time now and no one really complained, so I think it's safe to just keep things as is. A selection is essentially a list of object IDs, and if the object does not exist, we can handle that gracefully but still keep it in the selection object until the user updates it.

Basically I'm mostly worrying about the use-case when the operator is setting up a show, has stored a few groups for whatever purpose, and then there's a UI or server glitch and the operator just says "okay, I'm gonna restart the whole system". It would be preferable to keep the selection in that case.

All in all, if you are okay with it, I'll just remove the `selection` slice from the blacklist and we can merge this.

PR #153, opened 2026-03-26
Fix/trim commands from string

vasarhelyi:
This mini-PR trims command strings entered in the Properties->Messages panel before processing. I also removed checking of undefined strings as type indicated valid string. Hope it is OK.

### DIFF
--- a/src/flockwave/messages.ts
+++ b/src/flockwave/messages.ts
@@ -63,7 +63,8 @@ export function parseCommandFromString(string: string): {
   args: string[];
   kwds: Record<string, unknown>;
 } {
-  const parts = string.trim().length > 0 ? string.trim().split(/\s+/) : [''];
+  string = string.trim();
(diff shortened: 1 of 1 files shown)


src/flockwave/messages.ts
```diff
@@ -63,7 +63,7 @@ export function parseCommandFromString(string: string): {
   args: string[];
   kwds: Record<string, unknown>;
 } {
-  const parts = string && string.length > 0 ? string.split(/\s+/) : [''];
+  const parts = string.trim().length > 0 ? string.trim().split(/\s+/) : [''];
```
volfpeter: My only nit is we could `trim()` just once.

I think `string &&` in the original code was basically a nil check, I guess because this function is called from JS. I think it's fine to remove it.

volfpeter: Fixed.

PR #158, opened 2026-04-01
Count only assigned UAVs for show upload jobs

volfpeter:
The changes below affect only the show upload job:

- Status counters in the upload dialog ignore _unassigned UAVs that have a status somewhere in the job history_.
- The progress bar also ignores <same as above>
- The status of the "Upload show data" button in the control also ignores <same as above>

There is also a bugfix in this PR that fixes an edge-case in the upload saga (see my second commit in the PR).

How to reproduce the fixed bug:

- In `flockwave/operations.ts`, at the start of the `uploadDroneShow()` function, add some code that randomly throws an exception.
- Connect to the server and enable the virtual UAVs extension
- Load a show with plenty of UAVs (preferably >100)
- Start the show upload job for the entire fleet with auto-retry enabled
- Wait for an upload error
- Remove drone where the upload error occurred while the upload job is **still in progress**

Result before this PR: the upload job never is never completed, or if you removed multiple UAVs (with failed uploads) instead of just 1 while the upload is running, the application will hang.

Result after the fix: the upload job completes normally.

### DIFF
--- a/src/features/upload/saga.js
+++ b/src/features/upload/saga.js
@@ -81,21 +81,31 @@ function* runUploadWorker(chan, failed) {
     outcome = undefined;
     storedError = undefined;
 
+    let data = undefined;
+
     try {
       yield put(_notifyUploadOnUavStarted(uavId));
-      const data = selector ? yield select(selector, uavId) : undefined;
-      yield call(
-        executor,
-        { uavId, payload, data },
-        { onProgress: uploadProgressCallback }
-      );
-      outcome = 'success';
+      data = selector ? yield select(selector, uavId) : undefined;
     } catch (error) {
-      outcome = 'failure';
+      outcome = 'permanent-failure';
       storedError = error;
-    } finally {
-      if (yield cancelled() && !outcome) {
-        outcome = 'cancelled';
+    }
+
+    if (outcome === undefined) {
+      try {
+        yield call(
+          executor,
+          { uavId, payload, data },
+          { onProgress: uploadProgressCallback }
+        );
+        outcome = 'success';
+      } catch (error) {
+        outcome = 'failure';
+        storedError = error;
+      } finally {
+        if (yield cancelled() && !outcome) {
+          outcome = 'cancelled';
+        }
       }
     }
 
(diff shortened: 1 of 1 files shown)


src/features/upload/saga.js
```diff
@@ -81,21 +81,31 @@ function* runUploadWorker(chan, failed) {
...
-      outcome = 'success';
+      data = selector ? yield select(selector, uavId) : undefined;
     } catch (error) {
-      outcome = 'failure';
+      outcome = 'permanent-failure';
       storedError = error;
-    } finally {
-      if (yield cancelled() && !outcome) {
-        outcome = 'cancelled';
+    }
```
ntamas, line 92: This is really an edge case, but maybe we need to check for cancellation here as well? (i.e. maybe we need the same `finally` block here as the one we have below).

src/features/upload/selectors.ts
```diff
@@ -94,6 +95,28 @@ const getHistoryForSelectedJobType = createSelector(
...
+ *
+ * - `SHOW_UPLOAD_JOB_TYPE`: UAVs that don't participate in the mission don't
+ *   matter for example in upload status counts, cumulative upload status,
+ *   or enqueueing.
+ */
+const getFilteredUAVIdsForCurrentJobType = createSelector(
+  getSelectedJobTypeInUploadDialog,
+  _getFullMissionMapping,
+  (jobType, missionMapping): Set<Identifier> | undefined => {
+    return jobType === SHOW_UPLOAD_JOB_TYPE
```
ntamas, line 114: Note to self: maybe at some point we should extract this condition to a `jobTypeDependsOnMapping(jobType): boolean` helper function. Not necessarily now, but if we start seeing other places where this check becomes necessary then we definitely should.

volfpeter: This selector more or less does what you described. I think if the relevant UAV IDs depend on the job type, then eventually we'll need the UAV IDs in some form, so this selector can be used. But we'll see.

src/features/upload/saga.js
volfpeter: You're probably right. Could we maybe move the cancellation check *after* the `if (outcome === undefined) {}` block?

ntamas: On second thought it's probably not necessary to mocify anything here because the selector is synchronous so it's very unlikely that a cancellation event happens and _is handled_ while the selector is running.

PR #163, opened 2026-04-09
Update upload statuses after relevant state changes

volfpeter:
Changes:

- Refactored the `mission` slice: replaced a number of reducers with actions and added a `notifyUAVsInMissionMappingChanged()` reducer that can be used to update other slices when the mission mapping changes.
- Introduced an `'outdated'` UAV upload status. Outdated statuses hide previous statuses in the history during status aggregation.
- Mark UAV show upload statuses as outdated when the mission mapping changes. Only the statuses of affected UAVs get outdated (except when the mapping is fully replaced, but that usually happens after the mapping is cleared if I'm not mistaken, which obviously affects every UAV).
- Delete parameter upload statuses when the parameter list changes. In this case we always delete every status, I didn't want to refactor this `parameters` slice in this PR. We can improve it in the future.

### DIFF
--- a/src/features/mission/actions/actions.js
+++ b/src/features/mission/actions/actions.js
@@ -64,13 +64,13 @@ import { chooseUniqueId } from '~/utils/naming';
 import { createAsyncAction } from '~/utils/redux';
 import workers from '~/workers';
 
-import { JOB_TYPE } from './constants';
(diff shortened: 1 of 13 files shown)


ntamas (approved):
Very nice, thanks! The "outdated" status was a good idea that simplified lots of things.

PR #177, opened 2026-04-25
feat: add RTH method to show adapt params

vasarhelyi:
This PR adds RTH method to the show adapt trajectory params, similarly to already existing takeoff method. Functionality is already supported in the official latest version of the pro server, tested also, no need to update the server to support this new param.

One issue with the "organic takeoff + plain rth" combination: the plain RTH first brings drones to the end of their takeoff trajectories, but if organic takeoff is selected, this will be at the same altitude for all drones (as in organic takeoff they takeoff to the same height but in different times), possibly resulting in a min distance deviation if the single-layer takeoff is too dense. We might notify the user about this...

### DIFF
--- a/src/features/show-configurator/AdaptParametersForm.tsx
+++ b/src/features/show-configurator/AdaptParametersForm.tsx
@@ -7,7 +7,7 @@ import FormControl from '@mui/material/FormControl';
 import FormGroup from '@mui/material/FormGroup';
 import InputLabel from '@mui/material/InputLabel';
 import MenuItem from '@mui/material/MenuItem';
-import Select, { SelectChangeEvent } from '@mui/material/Select';
+import Select, { type SelectChangeEvent } from '@mui/material/Select';
 import { useCallback, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 
--- a/src/features/show-configurator/AdaptParametersForm.tsx
+++ b/src/features/show-configurator/AdaptParametersForm.tsx
@@ -20,8 +20,10 @@ import {
 } from '~/components/forms/fields';
 
 import {
+  RETURN_TO_HOME_METHODS,
   TAKEOFF_METHODS,
   type OptionalShowAdaptParameters,
+  type ReturnToHomeMethodType,
   type ShowAdaptParameters,
   type TakeoffMethodType,
 } from './actions';
@@ -37,6 +39,7 @@ const defaultAdaptParameters: ShowAdaptParameters = {
   verticalVelocity: 1.5,
   takeoffDuration: 0,
   takeoffMethod: 'layered',
+  returnToHomeMethod: 'smart',
 };
 
 const useStyles = makeStyles((theme) => ({
@@ -110,6 +113,9 @@ export function useAdaptParametersFormState(
       defaultAdaptParameters.takeoffDuration,
     takeoffMethod:
       defaultParameters?.takeoffMethod ?? defaultAdaptParameters.takeoffMethod,
+    returnToHomeMethod:
+      defaultParameters?.returnToHomeMethod ??
+      defaultAdaptParameters.returnToHomeMethod,
   });
   const [isValid, setIsValid] = useState(adaptParametersValid(parameters));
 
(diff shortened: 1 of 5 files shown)


src/features/show-configurator/AdaptParametersForm.tsx
```diff
@@ -212,6 +218,27 @@ export function useAdaptParametersFormState(
     },
     [onChange, parameters]
   );
+  const onReturnToHomeMethodChanged = useCallback(
+    (
+      event:
+        | React.ChangeEvent<{ value: string }>
+        | { target: { value: string } }
```
volfpeter, line 221: I should have a look at this sometime, it would be nice to simplify the typing. Same in `onTakeoffMethodChanged` (I introduced this for some reason).

volfpeter (approved):
Looks good.

Changelog needs update.

src/features/show-configurator/AdaptParametersForm.tsx
ntamas: This seems to do the trick for me:

```diff
diff --git a/src/features/show-configurator/AdaptParametersForm.tsx b/src/features/show-configurator/AdaptParametersForm.tsx
index 9f09197a..e647f2e3 100644
--- a/src/features/show-configurator/AdaptParametersForm.tsx
+++ b/src/features/show-configurator/AdaptParametersForm.tsx
@@ -7,7 +7,7 @@ import FormControl from '@mui/material/FormControl';
 import FormGroup from '@mui/material/FormGroup';
 import InputLabel from '@mui/material/InputLabel';
 import MenuItem from '@mui/material/MenuItem';
-import Select from '@mui/material/Select';
+import Select, { type SelectChangeEvent } from '@mui/material/Select';
 import { useCallback, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 
@@ -192,11 +192,7 @@ export function useAdaptParametersFormState(
     [onChange, parameters]
   );
   const onTakeoffMethodChanged = useCallback(
-    (
-      event:
-        | React.ChangeEvent<{ value: string }>
-        | { target: { value: string } }
-    ) => {
+    (event: SelectChangeEvent) => {
       const value: TakeoffMethodType = TAKEOFF_METHODS.includes(
         event.target.value as TakeoffMethodType
       )
```

ntamas: @volfpeter Any objections to what I proposed above? If not, I can make the change and then merge this PR.

src/features/show-configurator/AdaptParametersForm.tsx
```diff
@@ -352,6 +381,23 @@ const AdaptParametersForm = (props: Props): React.JSX.Element => {
...
+              </InputLabel>
+              <Select
+                disabled={disabled}
+                labelId='rth-method-label'
+                value={parameters.returnToHomeMethod}
+                onChange={onReturnToHomeMethodChanged}
+              >
+                {RETURN_TO_HOME_METHODS.map((value) => (
+                  <MenuItem key={value} value={value}>
+                    {t(`form.returnToHomeMethod.type.${value}`)}
```
ntamas, line 392: Since these translation keys are defined dynamically, we need to add them to `src/i18n/dynamic-keys.ts` in a comment to prevent the i18n tooling from removing them from `en.json` in the next iteration.

ntamas (changes requested):

src/features/show-configurator/AdaptParametersForm.tsx
volfpeter: Oh, I never noticed that type. Thanks! That's perfect!

src/features/show-configurator/AdaptParametersForm.tsx
volfpeter: Babel plugin :-1: Thanks though, I haven't seen that file yet. I guess we could also use `tt()` and prepare the keys in an object.

ntamas: I agree with your general sentiment re the Babel plugin, but unfortunately it's still there in the current setup so we need to cater for its needs. I'll take care of this, though, and then this can be merged.

PR #184, opened 2026-04-29
Feat/display num drones selected

### LINKED ISSUE #183: Display number of drones selected
Would be nice to display the number of drones selected. It is simply useful.

## Options:

1. around the UAV status widget's total number of drone count
2. on the basic view of the new selection group widget next to the icon with a single number
3. ??? any ideas ???

(I prefer option 2)

zarcell:
This branch adds a selected-items counter to the selection groups header button by introducing a new Redux selector (`getNumberOfSelectedItems`) and wiring it into `SelectionGroupsHeaderButton`, so the button now shows the current selected count (or `—` when none are selected).

Fixes #183.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added RTH method parameter to the show adaptation dialog box.
+- Added selected item count to the Selection groups header widget, matching the
+  typography of the neighboring UAV status widget.
 
 ## [2.13.2] - 2026-04-16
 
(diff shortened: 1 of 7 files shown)


vasarhelyi:
Some comments:

- please make the font type and size consistent with that of the UAV status widget next to it
- please add changes to CHANGELOG.md

otherwise it is nice, thanks! ... I am a bit worried about the changing width of the widget but it is the same with the UAV status widget so it is probably gonna be OK...

ntamas:
@zarcell As discussed on Discord: please add a "Selected objects" counter to the top of the tooltip that appears when hovering over the button, plus a divider line below it. Make sure that the text is localizable. I'll merge once this is one.

ntamas (approved):

PR #186, opened 2026-05-06
Fix: id ordering anomaly

### LINKED ISSUE #155: ID/sID ordering anomalies in the UAVs panel
The ID/sID columns in the UAVs panel are special and have somewhat peculiar behaviour. Would be nice to enhance their functionality along the following changes:

- rename the "Sort by mission IDs" button to "Mission mapping mode" 
- make the ID and sID columns independent of each other (right now you click on the second and IDs get ordered accoording to first column), i.e., enable ordering of both in both Normal and Mission views (but we can keep the order as is, namely, ID first in Normal view and sID first in Mission view)
- underscore the ID and sID columns as well if the UAVs are ordered according to them (now this is missing)
- perhaps remove the "Default" ordering from the "Sort by" list and replace with explicit ID and sID possibilities

and in the meantime keep the following functionality of the mapping mode:

- manual assignment and drag and drop reassignment of show IDs in edit mapping mode
- clear/hide missing mission IDs button
- and anything else that is there for a reason

zarcell:
Fixes: #155


ntamas:
@copilot resolve the merge conflicts in this pull request

ntamas:
@copilot please revert your merge commit

PR #112, opened 2025-12-08
feat(rtk): Add source preset management

zarcell:
This PR introduces a new dialog for managing RTK correction source presets, allowing users to create, edit, save, and delete presets for RTK connections.

PR #187, opened 2026-05-06
feat: Add source preset management to RTK modul

zarcell:
This was previously #112 but I converted it into a new PR since it was originated from the legacy fork I used at the time.


ntamas:
Okay, this is mostly fine; there are two remaining items that I intend to do separately:

* I don't like the `presetsRefreshTrigger` state variable in the `rtk` state slice; the time has probably come to store the list of RTK presets and the currently selected preset ID in the `rtk` state slice anyway so the refresh trigger will probably be gone and there will be a Redux action and a corresponding Redux listener to trigger a refresh.
* `src/flockwave/operations.ts` and `src/flockwave/queries.js` probably needs a general overhaul after so many years, but this is outside the topic of this PR.

PR #191, opened 2026-05-12
Warning if ground AMSL differs from altitudes

### LINKED ISSUE #185: Warning if ground AMSL differs from drone altitudes
Sometimes users forget to update the ground AMSL altitude before show that is a severe error as the show will be much higher or lower on the sky with all its consequences. Would be nice to have some kind of a warning around show upload if the AMSL value set for the ground level differs substantially from the average elevation of the drones.

zarcell:
Fixes #185.


src/features/uavs/selectors.ts
```diff
@@ -556,6 +570,93 @@ export const getActiveAndAwakeUAVIds = createSelector(
     })
 );
 
+const isUAVKnownToBeOnGround = (uav: StoredUAV): boolean => {
+  // Prefer explicit status codes from the UAV whenever they are available.
+  if (
+    uav.errors.includes(UAVErrorCode.ON_GROUND) ||
+    uav.errors.includes(UAVErrorCode.MOTORS_RUNNING_WHILE_ON_GROUND)
+  ) {
+    return true;
```
vasarhelyi, line 600: I would exclude drones that have some severe error. Basically any error code >= 128 (see https://docs.skybrush.io/public/skybrush-protocol-spec/latest/errors.html)

src/features/uavs/selectors.ts
```diff
@@ -556,6 +570,93 @@ export const getActiveAndAwakeUAVIds = createSelector(
...
+  // a known flight-related state.
+  if (
+    uav.errors.includes(UAVErrorCode.TAKEOFF) ||
+    uav.errors.includes(UAVErrorCode.LANDING) ||
+    uav.errors.includes(UAVErrorCode.RETURN_TO_HOME)
+  ) {
+    return false;
+  }
+
+  // Fall back to AHL only when the UAV does not provide a more explicit state.
```
vasarhelyi: What has the AHL got to do with our AMSL warning? It should not be related... Why do we need this extra possibility to return true? The on ground status codes are not enough?

src/features/show/stages.ts
```diff
@@ -99,8 +100,16 @@ const stages: Record<Stage, StageSpecification> = {
...
-      hasLoadedShowFile(state) && (hasShowOrigin(state) || isShowIndoor(state)),
+    evaluate(state) {
+      if (
+        !hasLoadedShowFile(state) ||
+        (!hasShowOrigin(state) && !isShowIndoor(state))
+      ) {
+        return false;
+      }
+
+      return selectGroundAMSLWarning(state) ? Status.WARNING : Status.SUCCESS;
```
vasarhelyi: I would _not_ disable further items in the show control panel, the warning is enough, this is not an error and there could be use cases when one wants to have a different AMSL on purpose.

vasarhelyi (changes requested):
Just a few small comments from a UI perspective, as an "outsider" to JS/TS code...

src/features/uavs/selectors.ts
zarcell: AHL is not part of the AMSL comparison; it only decides whether we treat the drone as on the ground for averaging AMSL when those explicit on-ground codes are absent. If we want stricter semantics we can remove this last-resort heuristic and rely only on those codes.

ntamas:
Started reviewing this and will continue tomorrow; two general comments that do not require action yet but I need to think about it:

* How do we handle this concept for indoor shows? Do we want to handle it for indoor shows at all?
* I think that the new setting and its related selectors should not include "AMSL" in the name; maybe simply "altitude warning threshold" is better than what we have in the PR. What we are doing is that we are comparing the average altitude of drones relative to _some global reference_ to the show's origin altitude, set by the user relative to the same global reference. It does not matter which global reference it is. We _happen_ to use AMSL (i.e. the reference is the mean sea level), but it could be relative to the Mount Everest or anything else. All that matters is that it's a threshold, in meters. So I'll probably rename stuff tomorrow but I'll get some sleep first.

ntamas:
I decided to allow stages dependent on `setupEnvironment` to be active even if `setupEnvironment` itself evaluates to a warning because of an AMSL reference mismatch. This is to account for cases when the average AMSL is _barely_ different from the configured reference AMSL, near the threshold. In such a situation, crew members lifting the drones manually while carrying them might trigger a change in the average AMSL that would result in dependent task being disabled and then re-enabled, and this would be quite annoying to the GCS operator.

PR #201, opened 2026-06-08
Convert queries to TS

volfpeter:
This is a more complex one...


src/flockwave/types/rtkSource.ts
```diff
@@ -0,0 +1,5 @@
+export type Response_XRTKSOURCE = {
```
ntamas: I think we have a `Response_RTKSOURCE` in `@skybrush/flockwave-spec` already. Not sure what's the best course of action with these messages.

Typically when we add a new message but we aren't committed to formalizing it yet in the spec we add it with an `X-` prefix (X meaning eXperimental). The server is configured not to validate incoming messages with this prefix against the official schema. Later on, when we are confident in the structure of the new message we remove the `X-` prefix and add official support for the message in the server.

I think we don't even have a handler for `X-RTK-SOURCE` in the server, only for `RTK-SOURCE`. I'm not sure where the `X-` prefix is stripped though. But I'm pretty confident that we could use `Response_RTKSOURCE` from `@skybrush/flockwave-spec`.

ntamas (changes requested):

src/flockwave/types/rtkSource.ts
volfpeter: Thanks, you right, `Response_RTKSOURCE` exists, I'll use that type instead.

volfpeter: I've updated the code. Since there's an `RTK-SOURCE` message in the spec, I removed the `X-` prefix too (there would be a TS error otherwise, because the `X-RTK-SOURCE` and `RTK-SOURCE` literal types don't overlap). I assume this is ok, but I couldn't test it locally.

ntamas: Tested locally, seems OK.

ntamas (approved):

PR #203, opened 2026-06-10
Convert more UAVDetailsDialog dependencies to TS

volfpeter:
There are multiple changes here:

- Flight log related typing improvements and extra validation
- Bugfix: only the last downloaded log was stored in Live, and it was downloaded, no matter which log the user requested (root cause: incorrect log type assumption, fixed with `JSON.stringify()` before text encoding)
- `UAVLogsPanel` is now in TS

@ntamas I should note that Live will now keep all downloaded logs in memory. Let me know if this could be a problem, and I'll add an LRU utility.

### DIFF
--- a/src/features/snackbar/types.ts
+++ b/src/features/snackbar/types.ts
@@ -14,7 +14,7 @@ export enum MessageSemantics {
 
 type ToastButton = {
   label: string;
-  action: Action;
+  action: Action | (() => void);
 };
 
 /**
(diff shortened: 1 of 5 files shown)


ntamas (approved):

ntamas:
LGTM. Can we add the LRU cache mechanism that you proposed, keeping, say, the last 50 downloaded logs? I doubt that anyone ever would need more than that.

volfpeter:
Created #204 for the LRU cache

PR #205, opened 2026-06-10
Convert PreflightStatusPanel to TS

volfpeter:
I removed the `PreflightCheckResult` enum to avoid conversions from the flockwave message. The types match, except Live adds an extra `unknown` option, which I didn't want to remove. Let me know if you'd prefer removing the `unknown` option or restoring the enum and converting in the `getPreflightStatus()` query.

### DIFF
--- a/src/components/MiniTable.tsx
+++ b/src/components/MiniTable.tsx
@@ -1,5 +1,5 @@
 import isNil from 'lodash-es/isNil';
-import PropTypes from 'prop-types';
+import type React from 'react';
 
 import { makeStyles } from '@skybrush/app-theme-mui';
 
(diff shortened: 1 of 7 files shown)


src/components/MiniTable.tsx
```diff
@@ -22,13 +22,17 @@ const useStyles = makeStyles((theme) => ({
...
     textAlign: 'right',
   },
-
-  separator: {},
 }));
 
 export const naText = <span className='muted'>—</span>;
 
-const MiniTable = ({ items }) => {
+export type MiniTableItem = string | [string, React.ReactNode];
```
ntamas, line 29: Is this correct? The original prop type was `PropTypes.oneOfType([PropTypes.string, PropTypes.arrayOf(PropTypes.node)])`, which would translate to `string | React.ReactNode[]`. But I guess the original intent was to say "anything that React accepts".

ntamas:
Looks fine to me; I don't mind removing the `PreflightCheckResult` enum now that we have a lower-level spec module (we did not have it at the time when this was introduced).

See my single comment above.

src/components/MiniTable.tsx
volfpeter: It's a string everywhere (`DockDetailsDialogSidebar` and `StatusSummaryMiniTable`). I think this was the intention.

If the item is a tuple, the first value is used as the `key`, which can't be a react node.

PR #207, opened 2026-06-15
Convert UAVDetailsDialog to TS

volfpeter:
Pending this PR in `@skybrush/mui-components`, which fixes the current type check error: https://github.com/skybrush-io/skybrush.js/pull/18

Besides the TS conversions, I've removed `SecondaryStatusLight` and the "secondary status" concept, because it wasn't used.


src/components/avatar/ComplexAvatar.tsx
```diff
@@ -83,25 +102,20 @@ const ComplexAvatar = ({
...
-}) => {
+  textSemantics = Status.INFO,
+}: ComplexAvatarProps) => {
   const classes = useStyles();
-
-  if (status === Status.INFO) {
-    status = Status.SUCCESS;
-  }
-
+  const effectiveStatus = status === Status.INFO ? Status.SUCCESS : status;
```
ntamas, line 121: I'm inclined to remove this override, but this does not affect the current PR.

ntamas (approved):

Issue #126, opened 2026-02-17
Persistent progress reporting of log downloads and calibrations


vasarhelyi:
Right now the progress reporting on the Tests and Logs dialog of Properties is local and not persistent. If user closes the properties dialog (e.g. with simply pressing Esc, or even with switching tabs), the progress disappears and cannot be revoked or displayed again. It is very annoying both for calibration tests and log downloads. Some additional related aspects:

- there is a new "calibrate compass" button on the UAVs header, that should start a progress reporter as well and we should be able to check individual status of drones being calibrated when opening the Properties only _after_ pressing that button
- when a single log download is started, it blocks communication with the drone and not even the log list can be retrieved later on, so not even the progress but the entire log list disappears on dialog close+open

volfpeter:
@vasarhelyi Task persistence is implemented in #210 

I started working on the "calibrate compass" button in UAVs view. It's a bit more complex than I originally thought, because this button performs a mass-operation with its own quirks. I may implement it in a separate PR.

The log download problem (your last point) feels like totally a different issue at first glance, but I'd need first need to debug what happens. I'd open a separate issue for that.

volfpeter:
Update from Tamás regarding the log download comment: you can't really do anything else with the drone during log download, it's an ArduPilot quirk, we can't really do anything about this in Live.

volfpeter:
The calibrate compass button part of the issue is implemented in #211

PR #210, opened 2026-06-22
Refactor: track log download and UAV test progress in redux

volfpeter:
Refs #126 

TODOs:

- [x] The Retry action is missing from the notification of the log download task. The operation can be retried from the log list, so this is a slight UX issue. I still want to re-add it in a clean way.
- [x] UAV tests that require user confirmation (pyro, motor) don't work, the confirmation button never appears. The feature might have broken earlier, must be fixed.


src/features/tasks/slice.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    _setTaskProgress(
+      state,
+      action: PayloadAction<TaskData & { progress: ProgressInfo }>
+    ) {
+      const { progress } = action.payload;
+      const key = getTaskKey(action.payload);
+      const entry = state[key];
+      if (entry) {
+        entry.progress = progress;
```
ntamas, line 41: Here we should probably set the status back to `'running'` if it is currently `'suspended'`. I don't think that the server sends an explicit message when it resumes working on a task after the user has provided the information or performed the action that the task needed (like: rotating the drone to the requested pose during accelerometer calibration), it just keeps on yielding progress messages.

src/features/tasks/actions/log-download-actions.ts
```diff
@@ -0,0 +1,88 @@
...
+    };
+
+    try {
+      const log = await messageHub.query.getFlightLog(uavId, logId, {
+        onProgress,
+      });
+      const hash = await logContents.write(log);
+      dispatch(_completeTask({ uavId, type, taskId, result: { hash } }));
+      showNotification({
+        message: `Log ${logId} of UAV ${uavId} downloaded successfully.`,
```
ntamas, line 66: Note to self (for the future): we need to be able to localize this eventually.

src/features/uavs/UAVLogsPanel.tsx
```diff
@@ -214,10 +38,16 @@ const UAVLogsPanel = memo(({ uavId }: UAVLogsPanelProps) => {
...
     () =>
-      uavId ? messageHub.query.getFlightLogList(uavId) : Promise.resolve({}),
+      uavId
+        ? messageHub.query.getFlightLogList(uavId)
+        : Promise.resolve(undefined),
     [messageHub, uavId]
   );
 
+  if (uavId === undefined) {
+    return <BackgroundHint text='No UAV selected' />;
```
ntamas, line 48: This need to be localizable.
Probably there are other similar strings in the codebase, maybe some of them are already localized in `en.json` so if there's an existing variant of this string in `en.json`, we could move that to a generic section in `en.json` and start using it everywhere.

src/features/uavs/UAVLogListItem.tsx
```diff
@@ -0,0 +1,172 @@
...
+        void writeBlobToFile(blob, filename);
+      }
+    }, [log]);
+
+    const onClick =
+      taskState?.status === 'running' ? undefined : log ? save : download;
+
+    const primaryParts: string[] = [
+      id,
+      timestamp === undefined ? 'Date unknown' : formatUnixTimestamp(timestamp),
```
ntamas, line 87: This string needs to be localized.

src/features/tasks/slice.ts
volfpeter: Thanks, that's a very good point. I fixed it. First I set task state unconditionally to `running`, but it's safer to only change the state of `suspended` tasks, so I added that condition.

src/features/tasks/actions/log-download-actions.ts
volfpeter: Yes, we need localization everywhere. When I started, I added some translations, but it made it harder to compare the code to the old implementation, so I reverted those changes.

When this PR is approved, I can stack another PR on top of it with just the i18n changes. That can be reviewed independently. There's another stacked PR (multi-UAV tasks for compass calibration). We could merge merge this one when the stacked ones are also reviewed and merged. At that point the feature will be complete.

src/features/uavs/UAVLogsPanel.tsx
volfpeter: Answered here: https://github.com/skybrush-io/live/pull/210#discussion_r3479718380

src/features/uavs/UAVLogListItem.tsx
volfpeter: Answered here: https://github.com/skybrush-io/live/pull/210#discussion_r3479718380

PR #217, opened 2026-07-03
More TS conversion

volfpeter:


src/features/weather/WeatherHeaderButton.tsx
```diff
@@ -25,16 +27,27 @@ import WeatherDetailsMiniList from './WeatherDetailsMiniList';
...
+};
+
+const WeatherHeaderButton = ({
+  badgeStatus,
+  sunrise,
+  sunset,
+}: WeatherHeaderButtonProps) => {
   /* Show sunset time if we are closer to the sunset than to the sunrise */
+  // eslint-disable-next-line @eslint-react/purity
   const now = new Date();
```
ntamas: We now have a `useCurrentDate()` hook that does this the correct (pure) way. Consider replacing this with the hook instead of disabling the warning.

ntamas (approved):
I'll fix the single comment that I had regarding the usage of `new Date()` in a React component and then merge the PR. Thanks!

PR #218, opened 2026-07-06
fix: keep fixed UAV ID and sID column order in list view

zarcell:
## Summary by CodeRabbit

* **UI Improvements**
  * UAV list columns now always display UAV ID first, followed by mission slot (sID).
  * List entries consistently show UAV ID as the primary label and mission slot as secondary information.
  * Updated mapping slot editor layout for improved alignment and sizing.
* **Documentation**
  * Clarified how mission-slot sorting affects grid views and dialogs, while list column order remains fixed.


src/views/uavs/MappingSlotEditorForList.jsx
```diff
@@ -12,7 +12,7 @@ import { getUAVIdForMappingSlotBeingEdited } from '~/features/mission/selectors'
 import { cancelMappingEditorSessionAtCurrentSlot } from '~/features/mission/slice';
 import { shouldOptimizeUIForTouch } from '~/features/settings/selectors';
 
-const WIDTH = 80;
```
ntamas, line 15: This changes the width of the editor widget. It was set exactly to encapsulate the ID _and_ the sID columns at the same time _and_ to place the text field directly over the "ID" column when the list was sorted by sID. Right now it's narrower and the text is misaligned. The text in the text editor _must_ overlap the text in the physical ID column in a pixel-perfect way.

src/views/uavs/SortAndFilterHeader.tsx
```diff
@@ -291,15 +290,8 @@ const ID_COLUMN: HeaderPart = {
...
-  ...COMMON_HEADER_TEXT_PARTS,
-]);
-
-/** List header used otherwise (UAV ID column is leftmost). */
-const LIST_HEADER_PARTS_UAV_FIRST: readonly HeaderPart[] = Object.freeze([
+/** List header: UAV ID column left, mission slot (sID) column right. */
+const LIST_HEADER_PARTS: readonly HeaderPart[] = Object.freeze([
   ID_COLUMN,
   SID_COLUMN,
   ...COMMON_HEADER_TEXT_PARTS,
```
ntamas: There's no need for a separate `COMMON_HEADER_TEXT_PARTS` segment if we are not messing with the order of the ID and sID columns. Just declare one array with the expected order of header parts and that's it.

src/views/uavs/UAVList.tsx
```diff
@@ -108,6 +108,9 @@ type ItemRendererOptions = {
   onSelectedItem: (item: string) => void;
   onStartEditing: (missionIndex: number) => void;
   selection: string[];
+};
+
+type GridItemRendererOptions = BaseItemRendererOptions & {
```
ntamas, line 115: Nitpicking: create a `ListItemRendererOptions` type as well even if it's just an alias for `BaseItemRendererOptions`, and then use it later in the `ListItemRenderer` props for consistency. (It bugged my eyes to see `BaseItemRendererOptions` there).

src/views/uavs/MappingSlotEditorForList.jsx
ntamas: Image

As a reference, this is how the editor appears in the current version. Note that it covers both columns -- the problem to fix is that the text editor containing physical ID being edited is now on the _right_ side but it should be on the _left_ if we never swap the placement of the ID and sID columns.

ntamas: Also make sure to test things with an ID that is 4 digits long. Things should look nice even if the ID has 4 digits. You might have to move the editor a bit to the left to make it pixel-perfect.

ntamas (changes requested):

vasarhelyi:
@zarcell if you change this functionality, please also update the relevant documentation in the `skybrush-live-doc` internal repo. Thanks!

PR #221, opened 2026-07-15
refactor: convert UploadStatusLights and UploadStatusRowHeader to TS

volfpeter:


src/features/upload/UploadStatusRowHeader.tsx
```diff
@@ -23,25 +19,28 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-const UploadStatusRowHeader = ({ label, onClick, uavIds, ...rest }) => {
+type UploadStatusRowHeaderProps = {
+  label: ReactNode;
+  onClick: (uavIds: string[]) => void;
```
ntamas, line 24: Can we  allow `onClick` to remain undefined? We might not always need a click handler for the row header.

ntamas (changes requested):
One minor comment; feel free to merge after it's fixed.

src/features/upload/UploadStatusRowHeader.tsx
volfpeter: The component is used in just one place and we always pass `onClick` as far as I can tell, so there's no point in keeping it optional. We can make it optional in the future if necessary, until then the stricter interface is cleaner in my opinion. I removed `...rest` for the same reason. I can restore `undefined` support, but I really don't see the point.

PR #222, opened 2026-07-16
Convert firmware update related components to TS

volfpeter:


src/features/firmware-update/FirmwareUpdateSetupDialog.tsx
```diff
@@ -50,30 +63,49 @@ const FirmwareUpdateTargetSelectorPresentation = selectableListOf(
...
   <AsyncGuard
     func={getTargets}
     errorMessage='Error while loading firmware update targets from server'
     loadingMessage='Retrieving firmware update targets...'
   >
     {(items) => (
-      <FirmwareUpdateTargetSelectorPresentation items={items} {...rest} />
+      <FirmwareUpdateTargetSelectorPresentation
+        items={items ?? []}
+        // TODO: do we need to pass the real target here?
```
ntamas: idk, probably not, it just happened to be the case that I forwarded all the remaining props to the presentation component "just in case". Does it work this way?

volfpeter: This is how `FirmwareUpdateTargetSelector` is used:

```ts
<FirmwareUpdateTargetSelector
  getTargets={getTargets}
  onChange={(_event, value) => setTarget(value)}
/>
```

So it never actually received a `value` props. That's why `value` is missing from the component's props and there's no `...rest` (we wouldn't need it anyway, a named prop would be cleaner).

Basically the question is whether this was a bug or was it intentional. In the first case we could pass a value, otherwise we should maybe make the prop optional in `SelectableListProps` (undefined would be a more explicit way of saying no selection).

ntamas: Not a bug and it makes sense to allow `undefined` in `SelectableListProps`. I'll fix this up and then merge the PR.

PR #223, opened 2026-07-17
Generalize mission planning, editing, uploading and progress monitoring to multiple UAVs

isti115:


src/features/mission/actions/actions.js
```diff
@@ -447,6 +450,40 @@ export const prepareMappingForSingleUAVMissionFromSelection =
...
+ * Thunk that assigns UAVs to the mission slots based on position matching.
+ */
+export const prepareMappingForMultiUAVMissionFromStartPositions =
+  (startPositions) => (dispatch) => {
+    dispatch(setMappingLength(startPositions.length));
+    dispatch(updateHomePositions(startPositions.map(toObjectFromScaledJSON)));
+    dispatch(recalculateMapping());
+  };
+
+export const changeMissionType = (newMissionType) => (dispatch, getState) => {
```
ntamas: I would probably call this `ensureMissionType()`. After all, it's not changing anything if the mission type is already the specified one, and it does not run side effects either in that case.

src/features/mission/actions/actions.js
```diff
@@ -447,6 +450,40 @@ export const prepareMappingForSingleUAVMissionFromSelection =
...
+
+    case MissionType.WAYPOINT: {
+      dispatch(clearMission());
+      break;
+    }
+
+    // No default
+  }
+
+  dispatch(setMissionType(newMissionType));
```
ntamas: Make `setMissionType` private-by-convention by prefixing its name with an underscore, and make sure that it is only called from here. I think that in all other cases we want mission type change requests to go through `changeMissionType` (or `ensureMissionType` if we decide to rename it).

src/features/mission/selectors/local.ts
```diff
@@ -959,41 +1129,44 @@ export const getNetMissionCompletionRatio: AppSelector<number> = createSelector(
...
  * from the local bounds and the local ratio.
  */
-export const getGlobalMissionCompletionRatio: AppSelector<number> =
-  createSelector(
-    getStartRatioOfPartialMission,
-    getEndRatioOfPartialMission,
-    getNetMissionCompletionRatio,
-    (start, end, localRatio) => start + (end - start) * localRatio
-  );
+export const getGlobalMissionCompletionRatio: AppSelector<
```
ntamas, line 1151: This function now takes a second argument, a mission index. However, it is used in `parameter-context.js` as a context handler, where it will _not_ receive a mission index. Is this handled gracefully?

(I found this by trying to convert `parameter-context.js` to TypeScript but then I backed out of it when it started to become too complex).

There might be other functions in `contextHandlers` in `parameter-context.js` that also suffer from this problem (haven't checked).

ntamas (changes requested):
I know that we are in "just do it" mode right now so these comments are only for the future:

* `UAVSelector` is getting a bit convoluted. First of all, we do not really use `UAVSelector` anywhere, we always use it through a `UAVSelectorWrapper` that wraps it so that the _real_ selector only appears as a popup (if my understanding is correct). Maybe `UAVSelectorWrapper` should become `UAVSelector` and `UAVSelector` should be renamed to `UAVSelectorGrid` or something like that.
* `UAVSelector` feels like two components in one for me right now, depending on the value of the `useMissionIds` prop (which is never set dynamically, it is always constant `true` or constant `false`). Maybe we could refactor `UAVSelector` to an "inner" component that gets a prepared array of `{ uavId, missionIndex }` objects, and wrap this inner component in variants that create the array in different ways (one for the normal UAV selection and one for the mission mapping based selection).

Haven't checked some files yet but I'm submitting the partial review anyway so we have something to talk about. Files still waiting for a review:

- [x] MissionOverviewList.jsx
- [x] MissionOverviewListItem.jsx
- [x] MissionOverviewPanelHeader.jsx
- [x] MissionOverviewPanelStatusBar.jsx
- [x] views/map/layers/mission-info.jsx

src/views/mission-editor/MissionOverviewListItem.tsx
```diff
@@ -246,13 +270,25 @@ const MissionOverviewListItem = ({
...
-            <Avatar className={isValid ? null : classes.error}>{avatar}</Avatar>
-          </ListItemAvatar>
-        )}
-        <ListItemText primary={primaryText} secondary={secondaryText} />
-      </ListItemButton>
+        <ListItemButton dense selected={selected} onClick={onClick}>
+          {avatar && (
+            <ListItemAvatar>
+              <Badge
+                badgeContent={item.participants
```
ntamas: We have a `formatIdsAndTruncateTrailingItems()` helper function in `~/utils/formatting` that deals with joining the IDs and trucating at a max length. Can we use that for sake of consistency? (The separator can be changed from the usual dot to a comma).

ntamas:
Finished checking all the files; review comments are above. The PR can be merged when the points above are addressed. Thanks a lot for your work on this!

isti115:
Thank you very much for the review, I'll start processing the requested changes. :+1:

src/views/mission-editor/MissionOverviewListItem.tsx
isti115: Sure! I've also moved this information to a chip instead, as the longer content didn't really fit into the badge anymore.

src/features/mission/actions/actions.js
isti115: Valid observation, rename done in 062dd1c9.

src/features/mission/selectors/local.ts
isti115: I believe (and testing shows) that while it will of course not return any meaningful results without the second parameter, it will not crash either. The entire concept of a global mission completion ratio with the partial re-plan / resume feature is very much defunct in its current state, as we haven't yet discussed / decided how to generalize it from a single UAV to multiple participants.

src/features/mission/actions/actions.js
isti115: I have thought about this as well while doing my self-review, but couldn't quickly come up with a clever way to break the cycle of recursive calls that would result from using `(change|ensure)MissionType` in the body of `clearLoadedShow`.

(There already is an import cycle between `src/features/show/actions` and `src/features/mission/actions`, which is kind of unfortunate on its own, but at least it's the sort of mutual dependence which doesn't lead to app-breaking runtime errors.)

Passing `void () => ensureMissionType(MissionType.UNKNOWN)` as the `onClearLoadedShow` event handler to `LoadShowFromFileButton` could be a simple workaround, but it just doesn't feel right to me. On the other hand if we import the internal `_clearLoadedShow` to use it directly in `ensureMissionType`, we didn't really make much progress by making `_setMissionType` private-by-convention.

One possible long-term architectural solution to this would be to let feature "modules" register their own mission types (or rather, _"operation"_ types, a terminology transition that is now long overdue to somewhat clean up the much overloaded _"mission"_ notion) and provide cleanup functions for them.

The best short-term way to avoid the infinite recursion may be to move the actual setting of the mission type above the cleanup invocations, in which case the `if (oldMissionType === newMissionType) { return }` guard would prevent the repeated calling of the functions. (This does feel like it leaves the store in a temporarily inconsistent state during the transition by entering the next "mode" before leaving the previous one properly, but might work well practically.)

isti115: In the meantime I have convinced myself to take the pragmatic approach and finding a way out of this by using a simple trick that works for now instead of spending more time thinking about the best possible theoretical solution. It can easily be improved later if we come up with something better, so I just added a `shouldClearMissionType` parameter to the cleanup functions that gets set to `false` when they are called from `ensureMissionType`...

ntamas (approved):

PR #231, opened 2026-08-25
Firmware consistency check job

### LINKED ISSUE #228: Firmware version comparison dialog
Similarly to the parameter comparison dialog, possibly reusing as much code as possible. The task is mostly the same: retrieve a `Map<string, T>` from the drone and report the differences between the most common values of `T` for each string key and the individual drones. The only difference is that `Map<string, T>` comes from a `UAV-VER` request instead of a `PRM-GET`. See the [UAV-VER spec](https://docs.skybrush.io/public/skybrush-protocol-spec/latest/messages/uav.html#_uav_ver_query_version_number_of_components) for more details.

volfpeter:
Closes #228

I had to move and slightly change a lot of utilities that were specific to the parameter consistency check job until now.

I didn't want more job-related utility packages in `features/`, so I moved/added all such code to `features/upload-support`. Maybe `upload-utils` would be a better name, let me know.

FWIW: the `upload` package name is quite misleading. Since we have a `tasks` package (relatively recent, for log downloads, compass calibration, etc), it may be time to rename `upload` to `jobs`, and `upload-support` to `job-support` (or `job-utils`). This change would affect lots of files, so this is not for this PR.

## Summary by CodeRabbit

* **New Features**
  * Added a firmware check that retrieves and displays vehicle firmware versions.
  * Added a firmware check option to the toolbox menu.
  * Added reusable upload setup dialogs, item lists, text parsing, and file import support.
  * Added value-consistency result views showing matching values, inconsistencies, and errors.

* **Improvements**
  * Standardized consistency and firmware check result handling.
  * Improved upload result panel configuration and dialog navigation.
  * Added English translations for firmware check features.


src/features/upload-support/value-consistency/analysis.ts
```diff
@@ -0,0 +1,97 @@
...
+  errors: Record<Identifier, string>;
+} => {
+  const distribution: ValueDistribution = Object.create(null);
+  const errors: Record<Identifier, string> = Object.create(null);
+
+  for (const [uavId, entry] of Object.entries(perUAVResults)) {
+    if (entry.type === 'success') {
+      for (const [name, value] of Object.entries(entry.result)) {
+        const valueAsString = String(value);
+        distribution[name] ??= Object.create(null);
```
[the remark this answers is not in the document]
volfpeter: I don't think this needs a fix.

ntamas:
> FWIW: the upload package name is quite misleading. 

Yes I know; part of the software evolution process. It was initially meant only for show uploads and we did not have any other operations yet that would have operated on multiple drones at once. Eventually it will have to be renamed, but not now.

src/features/upload-support/value-consistency/ValueConsistencyResultPanel.tsx
```diff
@@ -204,35 +214,45 @@ const ConsistencyCheckResultSummary = () => {
         : Status.SUCCESS;
   const message =
     errorCount > 0
-      ? t('consistencyCheck.result.errors', { count: errorCount })
+      ? messages.errors(t)
```
ntamas, line 219: Earlier version of the code fed the number of errors and/or inconsistencies into the message being translated and `react-i18next` was responsible for replacing the count tokens in the message (as well as providing the appropriate singular vs plural translation). How is this going to be handled here if we don't pass in the number of errors or inconsistencies?

(Okay to say that we are not going to support this any more, just wanted to ask in case it has just been forgotten accidentally).

src/features/firmware-check/actions.ts
```diff
@@ -0,0 +1,24 @@
...
+} from '~/features/upload/slice';
+import type { AppThunk } from '~/store/reducers';
+
+import { FIRMWARE_CHECK_JOB_TYPE } from './constants';
+
+/**
+ * Shows the upload dialog with a new firmware check job, or keeps the
+ * current job visible if a firmware check is already in progress.
+ */
+export const showFirmwareCheckDialog = (): AppThunk => (dispatch, getState) => {
```
ntamas, line 14: This seems like a pattern that is common enough to warrant a refactoring into a thunk factory that takes the job type and produces a function that could be used to open the dialog box "as is" if the current job type matches or to open the dialog box by overriding the job type if the job type does not match.

Maybe it could even be extended to handle jobs where we have an initial parameter setup step that is handled by a different dialog box.

Just an idea; feel free to ignore.

ntamas (approved):
LGTM; two minor comments only.

<sub>Approved in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/skybrush-io/live/pull/231)</sub>

ntamas:
Also, don't forget to update the changelog!

src/features/upload-support/value-consistency/ValueConsistencyResultPanel.tsx
volfpeter: That was already dead code since your refactor if I'm not mistaken: a0f2545

But I'll check the translations. As a result of this refactor I think I must to break up the `consistencyCheck` namespace, which I forgot to do.

src/features/firmware-check/actions.ts
volfpeter: I also thought about this. I'll have a look. If I remember correctly, it's different for parameter upload and consistency check (because of the setup dialog stage).

volfpeter:
> Also, don't forget to update the changelog!

Yes, I have a list of items that I need to add to the changelog. I'll go through my merged PRs and submit all in one commit.

src/features/upload-support/value-consistency/ValueConsistencyResultPanel.tsx
volfpeter: All good, no changes were required.

I added translations for firmware components as discussed. That change is in the stacked PR.

PR #240, opened 2026-09-22
fix: do not persist the open state of the collective RTH dialog

volfpeter:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved state persistence by excluding transient dialog activity from saved application data.
  * Preserved existing dialog state persistence behavior while preventing temporary dialog details from being retained.


ntamas:
@volfpeter Is this particularly annoying to you, more so than the similar behaviour of other dialogs?

I'm asking because if we decide to do this for the collective RTH dialog, then we should do it for all the other dialogs as well. But in practice I don't think it will bother anyone as most people don't close the app in the middle of a collective RTH planning process.

volfpeter:
@ntamas I don't think we persist the `open` state for other dialogs. At least not the ones I checked. In most cases the `dialog` key is excluded from persistence, which usually contains the `open` state. In this case there's no `dialog` wrapper in the state, that's why it seems (from the diff) that we're breaking a pattern. I thought about moving `open` under a 'dialog' key, but the state is already named "dialog" so I decided to skip the refactor. If you'd prefer to persist whether the dialog is open, feel free to close the PR. I can then change the behavior for other dialogs.

ntamas:
Hmm fair enough. I would prefer things to be consistent, so if we do not store the "open" states of all other dialogs, then we shouldn't store it here either (so the PR is okay).

Sounds like the right direction to go would be to move the state of this dialog under the `dialog` key and then we wouldn't have to deal with the exclusion of the `open` property separately. Is there anything else in the state slice that would not be okay under `dialog`?

